PR #97, opened 2012-12-12
Make write a lot faster

romainreuillon:
Querying for several writes at a time makes the write a lot faster as it hides the latency needed to get the response for each request.


mironabr:
This merge does not exists in the latest  mvn release 0.35.0.
Currently the upload speed using RemoteFileOutputStream is VERY slow!

dkocher:
This was reverted in 77924fd.

PR #141, opened 2014-08-14
Add proxy support for SocketClient/SSHClient

ziuchkovski:
This is a simple PR to add java.net.Proxy support to SocketClient.  It's in "works for me" status, tested against a SOCKS proxy.

I'm not sure if you want the added repetition on the connect() methods or not.  I just followed the way things were already laid out.

This -might- resolve https://github.com/shikhar/sshj/issues/10 as well, but I don't have an available HTTP proxy to test.


hierynomus:
Seems simple enough, merged.

PR #149, opened 2014-09-08
Add ability to change the protocol prefix string

bluekeyes:
I realize that this is a very niche feature, but I have a use case where it is required so perhaps others might as well. For context, I have an application that runs a modified Dropbear SSH server that follows the SSH protocol, but changes the protocol identification string to something else. I'm using SSHJ to communicate with both this server and standard SSH servers based on some configuration, so I needed a way to switch the protocol string without including multiple libraries.

Feel free to decline this and I'll just leave this change in my fork.

### DIFF
--- a/src/main/java/net/schmizz/sshj/Config.java
+++ b/src/main/java/net/schmizz/sshj/Config.java
@@ -88,6 +88,13 @@ public interface Config {
     String getVersion();
 
     /**
+     * Returns the protocol prefix string sent to and expected from the server during SSH connection initialization.
+     * This is almost always {@code "SSH"}, but may be different when connecting to embedded servers that follow the
+     * SSH protocol but give it a different name.
+     */
+    String getProtocolPrefix();
+
+    /**
      * Set the named factories for {@link Cipher}.
      *
      * @param cipherFactories a list of named factories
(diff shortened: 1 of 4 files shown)


hierynomus:
Code looks good, but I'm doubting whether to merge this. It is indeed not only a niche feature, it is actually not in the spec for as far as I know (correct me if I'm wrong). If I merge this, I'd at least want to see some unit tests that the defaults are still OK, and that it is possible to correctly change it.

A better solution would be to extract the protocol name and version into a separate class which you can pass to the Config object. Because now you changed the prefix, but the version is still "fixed", if we do this, we might as well also extract the version, but then I would actually expect a `ProtocolVersion` class to be injected into the config. The `ProtocolVersion` would then also be able to provide some kind of "strategy" for the protocol to handle. In that way we could also support SSHv1 (the old protocol), which I've run into once before with SSHj.

Currently I will not merge it, given above thoughts, I'd be more inclined to correctly implement this, and then this PR is too simplistic.

bluekeyes:
That makes sense, the current implementation was just the minimum needed for my use. If I have some time, I'll look at making this more general and better tested. If the result is not that invasive, the flexibility might be nice, but otherwise I'll just keep this minimal fork for my own use.

Regarding the spec, this actually contradicts [section 4.2](https://tools.ietf.org/html/rfc4253#section-4.2) of the transport layer protocol, but that's expected: my "new" protocol is by definition not SSH, it just behaves like SSH in all ways except for the protocol string.

hierynomus:
Indeed it does... I was just reading it again for another issue and it is
contradictory... But extracting it would allow support for sshv1 to be
easily added (which is also not in the spec ;)).

bluekeyes:
I'm going to close this - the project I added it for changed direction and I don't need this feature anymore. Feel free to reopen if it would be useful for SSHv1 support (and if that's something you actually want to develop further).

PR #156, opened 2014-11-11
fixed block sizes for hmac-sha2-256 and hmac-sha2-512

Boris-de:
These MACs where added in fdb891b842 but they didn't work for me after disabling md5&sha1. The server simply didn't respond anymore eventually resulting in a timeout. I debugged this and found two issues with the current code:

* both MACs have to use larger block/digest sizes, the previous values were the same that SHA1 had, but SHA2 needs 32 or 64 byte
* Additionally the KeyExchanger would only generate a 20 byte key when using diffie-hellman with sha1 as a KEX. For SHA2 larger keys are used, so the first message with a mac resulted in a "Corrupted MAC on input" error on the server. So the keys had to be resized like its already done for the Ciphers.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -26,7 +26,7 @@ signatures::
   `ssh-rsa`, `ssh-dss`
 
 mac::
-  `hmac-md5`, `hmac-md5-96`, `hmac-sha1`, `hmac-sha1-96`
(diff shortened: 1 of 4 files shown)


hierynomus:
Thx indeed.. Just catching up on issues and PRs :)

hierynomus:
It's merged already, but could you also provide some unit tests for this?

Boris-de:
Great, thanks for merging. I created a new PR with some tests for the MACs, see #164

Issue #175, opened 2015-02-11
Port forward to unavailable destination cause dead lock on client side


kondratovich:
Hello, [source](https://github.com/hierynomus/sshj/blob/master/src/main/java/net/schmizz/sshj/connection/channel/direct/LocalPortForwarder.java#L133).

In case, destination in parameters is not available, `open channel` fails with exception

```
Opening direct-tcpip channel failed: Connection refused
	at net.schmizz.sshj.connection.channel.direct.AbstractDirectChannel.gotOpenFailure(AbstractDirectChannel.java:64)
	at net.schmizz.sshj.connection.channel.direct.AbstractDirectChannel.gotUnknown(AbstractDirectChannel.java:89)
	at net.schmizz.sshj.connection.channel.AbstractChannel.handle(AbstractChannel.java:193)
	at net.schmizz.sshj.connection.ConnectionImpl.handle(ConnectionImpl.java:128)
	at net.schmizz.sshj.transport.TransportImpl.handle(TransportImpl.java:455)
	at net.schmizz.sshj.transport.Decoder.decode(Decoder.java:107)
	at net.schmizz.sshj.transport.Decoder.received(Decoder.java:175)
	at net.schmizz.sshj.transport.Reader.run(Reader.java:61)
```

`DirectTCPIPChannel` is not started, socket monitor is not activated and socket is never closed. Client is locked forever.

hierynomus:
Thx for finding this! I'd love a PR to fix :) Else I'll look into it soon.

PR #176, opened 2015-02-12
Local port forwarder: Catch open failure

kondratovich:
#175


hierynomus:
Would it be possible to add a (few) unit test(s)? To prove that it works correctly.

hierynomus:
Fixed via e6c7c176643a1605f7bf38ae9d90cdab31545d00

Issue #180, opened 2015-02-24
Invalid write packet length


dkocher:
When reading from a source buffer with an offset.

hierynomus:
Fixed with merging #181.

PR #181, opened 2015-02-24
Fix length field for SSH_FXP_WRITE packets.

dkocher:
Proposed fix for #180.

### DIFF
--- a/src/main/java/net/schmizz/sshj/sftp/RemoteFile.java
+++ b/src/main/java/net/schmizz/sshj/sftp/RemoteFile.java
@@ -88,7 +88,7 @@ public class RemoteFile
             throws IOException {
         return requester.request(newRequest(PacketType.WRITE)
                                          .putUInt64(fileOffset)
-                                         .putUInt32(len - off)
+                                         .putUInt32(len)
                                          .putRawBytes(data, off, len)
         );
     }


hierynomus:
Wondering whether the code @shikhar wrote is actually correct here. According to the spec there is no length field, i.e.

```
  Writing to a file is achieved using the SSH_FXP_WRITE message, which
   has the following format:

   	uint32     id
   	string     handle
   	uint64     offset
   	string     data

   where `id' is a request identifier, `handle' is a file handle
```

(source: http://www.openssh.com/txt/draft-ietf-secsh-filexfer-02.txt)

dkocher:
I think the `length` field is just omitted in the spec because it is described in the *General Packet Format* section.

hierynomus:
That field is actually handled in the [SFTPEngine#transmit](https://github.com/hierynomus/sshj/blob/master/src/main/java/net/schmizz/sshj/sftp/SFTPEngine.java#L267) method. That ensures that it transmits the `packet length` first, and then the payload (packet type, etc). So I do think that the length field should not be there (at least according to the spec)

dkocher:
I will run some interoperability tests with OpenSSH if omitting the field is fine then.

dkocher:
I have errors when omitting the length in the before the raw bytes in the `WRITE` packet.

```
net.schmizz.sshj.sftp.SFTPException: EOF while reading packet
	at net.schmizz.sshj.sftp.PacketReader.readIntoBuffer(PacketReader.java:54)
	at net.schmizz.sshj.sftp.PacketReader.getPacketLength(PacketReader.java:59)
	at net.schmizz.sshj.sftp.PacketReader.readPacket(PacketReader.java:75)
	at net.schmizz.sshj.sftp.PacketReader.run(PacketReader.java:87)
```

hierynomus:
I'll merge this PR, as it indeed fixes the bug... Still wondering though why we need to length field in the SSH_FXP_WRITE packet. Opened #187 for this.

PR #189, opened 2015-04-13
Fix Android compatibility

### LINKED ISSUE #188: #136 / #134 (version 0.10) broke android support
ECDSA support is not found in Android's BouncyCastle implementation. Dropping in a different ECDSA provider does not work (http://rtyley.github.io/spongycastle/) because there is a strong dependency on the BouncyCastle EC classes in `KeyType`.

I might try to tackle this sometime soon.

yincrash:
Fixes #188 by removing ECDSA signatures in the AndroidConfig.
Add SpongyCastle as the default security provider.

### DIFF
--- a/src/main/java/net/schmizz/sshj/AndroidConfig.java
+++ b/src/main/java/net/schmizz/sshj/AndroidConfig.java
@@ -15,12 +15,42 @@
  */
 package net.schmizz.sshj;
 
+import java.security.Provider;
+import java.security.Security;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import net.schmizz.sshj.common.SecurityUtils;
+import net.schmizz.sshj.signature.SignatureDSA;
+import net.schmizz.sshj.signature.SignatureRSA;
 import net.schmizz.sshj.transport.random.JCERandom;
 import net.schmizz.sshj.transport.random.SingletonRandomFactory;
 
 public class AndroidConfig
         extends DefaultConfig {
 
+    static {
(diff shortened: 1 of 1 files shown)


hierynomus:
This is not complete. If somebody would have called upon `SecurityUtils` before loading the `AndroidConfig`, it would break.

yincrash:
Whoops. I'll fix it and add a test.

hierynomus:
Fixed this.

willemw12:
Is this fixed? The changes have not been merged. There is no:

    .... SecurityUtils.setSecurityProvider("SC"); ...

in AndroidConfig.java.

hierynomus:
Correct, The call to `SecurityUtils.registerSecurityProvider("org.spongycastle....");` in `AndroidConfig` does that for you. See https://github.com/hierynomus/sshj/blob/master/src/main/java/net/schmizz/sshj/common/SecurityUtils.java#L73

So yes, this should be fixed. Unless you've found an issue :)

willemw12:
I could only make SSHJ+SpongyCastle work after temporarily changing isBouncyCastleRegistered() in SecurityUtils.java, from:

    return BOUNCY_CASTLE.equals(securityProvider);

into:

    return BOUNCY_CASTLE.equals(securityProvider) || "SC".equals(securityProvider);

hierynomus:
@willemw12 Can you create a new issue for that?

willemw12:
@hierynomus: See https://github.com/hierynomus/sshj/issues/308.

hierynomus:
Thanks @willemw12 Will fix it.

PR #195, opened 2015-05-18
Add support for "gssapi-with-mic" authentication (Kerberos)

bluekeyes:
This is working with an internal (soon to be open-source) library we have, but there might still be problems with certain edge cases. Not sure what your testing policy is, but I can look into writing some tests if the MINA SSHD server supports "gssapi-with-mic" authentication as well.


bluekeyes:
Relevant RFC for reference: https://www.ietf.org/rfc/rfc4462.txt

hierynomus:
Seems like `gssapi-with-mic` is supported by Apache Mina (https://issues.apache.org/jira/browse/SSHD-111). Would be nice to have some tests... Also would be good start to test all the other new key exchange mechanisms as well if they're supported by Mina.

dkocher:
I think this patch is quite awesome. Would love to see a sample how to use it with a login context.

hierynomus:
Definitely agreed that the patch is very cool.

bluekeyes:
@dkocher Here's an example of how we set it up for Kerberos. Unfortunately, it requires setting several system properties and generating a file. I need to do some more research to see if it can be done entirely in code.

```java
String username;
String kerberosRealm;
String kdcHostname;

/* 
 * See http://docs.oracle.com/javase/7/docs/technotes/guides/security/jgss/tutorials/LoginConfigFile.html
 * 
 * For Kerberos, the file contains:
 *
 * Krb5LoginContext {
 *   com.sun.security.auth.module.Krb5LoginModule required refreshKrb5Config=true useTicketCache=true;
 * };
 *
 */
File jaasConfigFile;

// Required by Krb5LoginContext
System.setProperty("java.security.krb5.realm", kerberosRealm);
System.setProperty("java.security.krb5.kdc", kdcHostname);
System.setProperty("java.security.auth.login.config", jassConfigFile.toString());

LoginContext lc = null;
try {
    lc = new LoginContext("Krb5LoginContext");
    lc.login();
} catch (LoginException e) {
    throw new UserAuthException(e);
}

Oid krb5Oid;
try {
    // as defined by the Kerberos specification
    krb5Oid = new Oid("1.2.840.113554.1.2.2");
} catch (GSSException e) {
    throw new UserAuthException("Failed to create Kerberos OID", e);
}

SSHClient client = new SSHClient();
client.authGssApiWithMic(username, lc, krb5Oid);
```

bluekeyes:
After a lot of mocking, I got a simple authentication test working. Any other comments?

Also, is there any timeline on when this could merge? I have a project I'm working to open source that depends on this feature being available in a public SSHJ release.

hierynomus:
Looking at it now :+1:

PR #196, opened 2015-05-28
bugfix: match complete host instead of contains on the hoststring

Boris-de:
The SimpleEntry currently matches the hostname of the connection against
the complete hoststring of the entry. This way substrings also match, so
for example "10.0.0.1" matches on an entry for "10.0.0.10", resulting in
a host-key-changed message if the key differs which is usually does.


hierynomus:
Thanks!

PR #200, opened 2015-06-16
Add option for client to verify negotiated key exchange algorithms.

dkocher:


src/main/java/net/schmizz/sshj/sftp/SFTPEngine.java
```diff
@@ -59,7 +59,12 @@ public SFTPEngine(SessionFactory ssh)
 
     public SFTPEngine(SessionFactory ssh, String pathSep)
             throws SSHException {
-        sub = ssh.startSession().startSubsystem("sftp");
+        this(ssh, pathSep, "sftp");
+    }
+
+    public SFTPEngine(SessionFactory ssh, String pathSep, String subsystemName)
```
hierynomus: Part of this PR? Seems unrelated at first sight.

dkocher: Yes, completly unrelated. I should work on branches that can be pulled.

hierynomus:
Definitely like the idea! If you can remove the unrelated commit ;)

src/main/java/net/schmizz/sshj/sftp/SFTPEngine.java
hierynomus: Let me know when you've fixed that :)

PR #208, opened 2015-08-12
SCP : limit the used bandwidth

lguerin:
Hello,

I have a strong requirement to limit usage of bandwidth at application level when I upload files to a remote server. 
According to the scp man page, I have added a '-l XXX' arg to the SCPEngine class.
Then, to keep the fluent aspect of your API, this new arg can (optionally) be set as below:

```java
// Limit upload transfert at 128Ko/s
ssh.newSCPFileTransfer().bandwidthLimit(1024).upload(new FileSystemFile(src), "/tmp/");
```

A test case is provided into SCPFileTransferTest for upload and download features.
Thanks.

### DIFF
--- a/src/main/java/net/schmizz/sshj/xfer/scp/SCPEngine.java
+++ b/src/main/java/net/schmizz/sshj/xfer/scp/SCPEngine.java
@@ -28,6 +28,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.LinkedList;
 import java.util.List;
 
 /** @see <a href="http://blogs.sun.com/janp/entry/how_the_scp_protocol_works">SCP Protocol</a> */
@@ -39,17 +40,31 @@ class SCPEngine {
         RECURSIVE('r'),
         VERBOSE('v'),
         PRESERVE_TIMES('p'),
-        QUIET('q');
+        QUIET('q'),
+        LIMIT('l', "");
 
         private final char a;
+        private String v;
 
         private Arg(char a) {
             this.a = a;
         }
+        private Arg(char a, String v) {
+            this.a = a;
+            this.v = v;
+        }
+
+        public void setValue(String v) {
+            this.v = v;
+        }
 
         @Override
         public String toString() {
-            return "-" + a;
+            String arg = "-" + a;
+            if (v != null && v.length() > 0) {
+                arg = arg + v;
+            }
+            return arg;
         }
     }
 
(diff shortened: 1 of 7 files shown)


hierynomus:
Hi! Thanks for the PR. Unfortunately, Travis claims that one test is failing. Can you have a look and fix it?

lguerin:
Hi,
I don't really know why the first build has failed, whereas the tests was ok on my local machine and with my travis instance: https://travis-ci.org/lguerin/sshj/builds/75394840
I just have change the logging level for tests in gradle file to get more informations about the exception cause if the error occurs again.

src/main/java/net/schmizz/sshj/xfer/scp/SCPFileTransfer.java
```diff
@@ -28,18 +28,23 @@
         extends AbstractFileTransfer
         implements FileTransfer {
 
+    /** Default bandwidth limit for SCP transfert in Kbit/s (-1 means unlimited) */
```
hierynomus: typo in comment (transfert). Also is it kilobit, or kilobyte? The difference is a factor 8.

lguerin:
Sorry for the typo. It's kilobit/s, according to the scp man page.

src/main/java/net/schmizz/sshj/xfer/scp/SCPEngine.java
```diff
@@ -39,17 +40,31 @@
         RECURSIVE('r'),
         VERBOSE('v'),
         PRESERVE_TIMES('p'),
-        QUIET('q');
+        QUIET('q'),
+        LIMIT('l', "");
 
         private final char a;
+        private String v;
```
hierynomus: Why the extra 'v' here? It is not used currently. The `LIMIT` _seems_ to use it, but as it is an enum it is always the empty string.

src/test/java/net/schmizz/sshj/xfer/scp/SCPFileTransferTest.java
```diff
@@ -0,0 +1,82 @@
...
+    
+    @After
+    public void cleanup() {
+        if (targetFile.exists()) {
+            targetFile.delete();
+        }
+    }
+
+    @Test
+    public void should_SCP_Upload_File() throws IOException {
```
hierynomus: I like the tests, but could you name them without the underscores and with just camelcasing?

src/main/java/net/schmizz/sshj/xfer/scp/SCPEngine.java
lguerin: Some scp options needs a key/value structure. The value is initialized with an empty string but can be set through the helper class SCPArguments. This the way I have used to set the value of the LIMIT arg.

src/test/java/net/schmizz/sshj/xfer/scp/SCPFileTransferTest.java
lguerin: Done

src/main/java/net/schmizz/sshj/xfer/scp/SCPEngine.java
hierynomus: Ahh but this is even more evil, as this is an enum, setting the value for 1 process means that you're setting it (inadvertently) for all processes. Because there is only 1 instance of the same enum key in a JVM. So please refactor that and remove the `setValue(..)`

hierynomus:
Thanks! Looks good now I think. Will do some testing on it, and then merge it in next week.

lguerin:
Nice, thank you.

hierynomus:
Merged, thx!

PR #210, opened 2015-08-17
Fix issue 209.

dkocher:
Please review.


hierynomus:
Hi David, this qualifies for a small testcase :) On first sight the fix looks good, but I'd rather see some tests

PR #211, opened 2015-08-17
Skip blank lines when detecting key formats

bluekeyes:
I found a set of systems (CI nodes) where the private key files started with a blank line, which broke tests using SSHJ. We fixed the key files, but since OpenSSH, git, and all the other applications we use worked with these files, it seemed like an SSHJ bug.

Also add some basic tests for key formats.


hierynomus:
Thanks! Looks great.

Issue #247, opened 2016-05-17
Allow mocking of SCPUploadClient and SCPDownloadClient


joprsal:
Currently, there is no way of mocking the two final classes using Mockito (because [Mockito cannot mock final classes](https://github.com/mockito/mockito/wiki/FAQ#what-are-the-limitations-of-mockito)).

One solution would be to extract and use interfaces rather than final classes.

hierynomus:
This is fixed with PR #250

PR #248, opened 2016-05-17
Fix for issue #247: Allow mocking of SCPUploadClient and SCPDownloadClient

joprsal:
This change is to fix issue #247 

Summary of change:
- implementation classes renamed
  * SCPDownloadClient --> SCPDownloadClientImpl
  * SCPUploadClient --> SCPUploadClientImpl
- extracted interfaces (SCPDownloadClient & SCPUploadClient)
- made SCPFileTransfer use interfaces instead of implementation classes

Backwards compatibility note:
Since the new interfaces' names and public methods are exactly the same, existing code should not be affected.


hierynomus:
Unfortunately that is not true. This requires that all code using sshj is
recompiled.
Op 17 mei 2016 14:51 schreef "Juraj Oprsal" <notifications@github.com>:

joprsal:
Ok, my bad, you are right.

Would you accept a change to make those two classes non-final? Such change would not require existing code to be recompiled.

hierynomus:
Agreed, let's make them non-final.

joprsal:
Closing this pull request as rejected. Will raise a new one with agreed change.

PR #253, opened 2016-07-04
Ensure that same name subdirs aren't omitted when doing SFTP

hierynomus:
@dkocher Could you verify whether CyberDuck relies on the following behaviour and what your preference would be...

Currently when an SFTP copy operation starts, we do a check to see whether the source is a directory, and whether the given destination is a directory of the same name. In that case we copy the contents of the source into the destination. If the names however do not match, we upload the source directory itself into the destination (creating a subdir).

I know we've worked around this in Overthere already (see: https://github.com/xebialabs/overthere/blob/master/src/main/java/com/xebialabs/overthere/ssh/SshSftpFile.java#L143), as the other protocols behave differently. So I was wondering whether we should remove this check, or maybe make it configurable to always upload the contents of the source to the destination.


dkocher:
@hierynomus We do not depend on `SFTPFileTransfer ` in Cyberduck therefore I have no strong opinion on this.

PR #262, opened 2016-08-13
Added PKCS5 key file support (without requiring BouncyCastle)

joval:
I'm doing some experimentation with this SSH client, to see if we might leverage it in our project.  However, we don't want to add a hard requirement for BouncyCastle just to perform public key authentication.

Therefore, I've implemented a PKCS5KeyFile class that can parse PKCS5 key files.  I've also updated the test resource for pkcs8 (as it was actually a pkcs5 file), and updated the test classes as well.

Additional minor modifications I've made permit interoperability with older versions of SLF4J (back to 1.6), and the addition of a few methods to the OpenSSHKnownHosts.HostEntry interface (to make it possible to get the entry type and fingerprint string).

I plan to add a few more features in the coming days.


hierynomus:
Hi Joval,

Thanks for this contribution. As a suggestion, could you split different things up into multiple PRs?

The older SLF4J support for instance had no real place in this PR, neither did the additions to the HostEntry interface. 

Still a great PR though, awaiting your other features!

joval:
Yeah, sorry, I'll try to group changes more logically in the future.  Thanks for accepting the changes!

PR #263, opened 2016-08-16
Modifications for working with SSH gateways

joval:
In the spirit of keeping PRs functionally-targeted, submitted for your approval are a few changes pertaining to working with SSH gateways and proxies.

First, I re-ordered the connect methods in SocketClient.java, so that all the String-based and InetAddress-based methods would be grouped together (so that I wouldn't have to jump around in the file when working on them).  Then, I eliminated all use of InetAddress.getByName in the String-based connect methods, because when you're connecting through a gateway or Proxy you may not actually have the ability to resolve the DNS name of the target host (but the Proxy or gateway will).

I also added some lightweight management of LocalPortForwarders to the SSHClient class, because when the client is closed, I want any LocalPortForwarders that are using the client's transport to also be closed.  This makes it possible to have a stateless SocketFactory that connects Sockets through an SSH gateway (SocketFactories being the preferred method of implementing connections through gateways in sshj).

I considered having the SSHClient class actually manage the ServerSockets and listener threads associated with LocalPortForwarders (as JSch does), but thought you might not want that.  So for now I'm doing all that in a SocketFactory class that sits outside of the sshj project.

### DIFF
--- a/src/main/java/net/schmizz/sshj/SSHClient.java
+++ b/src/main/java/net/schmizz/sshj/SSHClient.java
@@ -139,7 +139,7 @@ public class SSHClient
     /** {@code ssh-connection} service */
     protected final Connection conn;
 
-    private List<LocalPortForwarder> forwarders;
+    private final List<LocalPortForwarder> forwarders = new ArrayList<LocalPortForwarder>();
 
     /** Default constructor. Initializes this object using {@link DefaultConfig}. */
     public SSHClient() {
@@ -433,15 +433,14 @@ public class SSHClient
     @Override
     public void disconnect()
             throws IOException {
-        if (forwarders != null) {
-            for (LocalPortForwarder forwarder : forwarders) {
-                try {
-                    forwarder.close();
-                } catch (IOException e) {
-                    log.warn("Error closing forwarder", e);
-                }
+        for (LocalPortForwarder forwarder : forwarders) {
+            try {
+                forwarder.close();
+            } catch (IOException e) {
+                log.warn("Error closing forwarder", e);
             }
         }
+        forwarders.clear();
         trans.disconnect();
         super.disconnect();
     }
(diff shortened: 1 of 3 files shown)


src/main/java/net/schmizz/sshj/SSHClient.java
```diff
@@ -431,6 +433,15 @@ public void authGssApiWithMic(String username, LoginContext context, Oid support
...
     public void disconnect()
             throws IOException {
+        if (forwarders != null) {
+            for (LocalPortForwarder forwarder : forwarders) {
+                try {
+                    forwarder.close();
+                } catch (IOException e) {
+                    log.warn("Error closing forwarder", e);
+                }
+            }
```
hierynomus: clear the forwarders list after the loop?

src/main/java/net/schmizz/sshj/SSHClient.java
```diff
@@ -648,7 +659,12 @@ public void loadKnownHosts(File location)
      */
     public LocalPortForwarder newLocalPortForwarder(LocalPortForwarder.Parameters parameters,
                                                     ServerSocket serverSocket) {
-        return new LocalPortForwarder(conn, parameters, serverSocket);
+        if (forwarders == null) {
```
hierynomus: If the list is initialized this check is unnecessary.

src/main/java/net/schmizz/sshj/SSHClient.java
```diff
@@ -139,6 +139,8 @@
     /** {@code ssh-connection} service */
     protected final Connection conn;
 
+    private List<LocalPortForwarder> forwarders;
```
hierynomus: Make final and initialize with `new ArrayList<>()`? This prevents the excessive null checks.

hierynomus:
Tests are now failing, due to the moves of the methods it is somewhat hard to see why 😉

joval:
I see that ... server.getHost() at SshFixture.java line 104 returns null?  But what did I do that would cause that?

hierynomus:
It probably was already returning null, but we were able to handle that?

joval:
Aha!

From InetAddress.getByName(String):
"If the host is null then an InetAddress representing an address of the loopback interface is returned. See RFC 3330 section 2 and RFC 2373 section 2.5.3."

SshServer.getHost() was always returning null, but this previously became an InetAddress to the loopback interface.

Yes, how do you want to address this?  To me it's counter-intuitive to assume null == loopback.

hierynomus:
Well I think there are 2 valid reasons for keeping it like this:

1. It's in the RFC's, so this is specced behaviour.
2. Changing it might cause problems for people using the library, as it would be backwards incompatible.

joval:
OK, I'll check for a null hostname and set it to a loopback to emulate the old behavior.

PR #265, opened 2016-08-19
Minor modifications to PKCS5KeyFile

joval:
Just wanted to get this out of the way before tackling the logger factory issue.


hierynomus:
I would like to see an added testcase ;) Prevents stuff from breaking again when refactoring

hierynomus:
Also [codacy](https://www.codacy.com/app/jeroen_2/sshj/pullRequest?prid=323080#newIssuesView) shows that you've introduced 2 new code issues with this PR

joval:
I'm not sure what to make of the two codacy issues:

1) I put the new field up with all the other fields, following your convention of putting the factory inner class definition up top.  Although, I'll move the new inner exception classes to the top since that also appears to be the convention in the code-base.

2) I moved the decryption to its own method, and actually (I thought) simplified it!  Now it complains?

I will add a test case.

hierynomus:
I've just added codacy, so there might still be some things in the ruleset that can be tweaked ;) I agree with keeping the static factory classes on top.

As for the NPath complexity, it complains that it is still too high. Reduction can normally be achieved by extracting into different methods for different paths :)

src/main/java/net/schmizz/sshj/userauth/keyprovider/PKCS5KeyFile.java
```diff
@@ -126,85 +146,56 @@ protected KeyPair readKeyPair()
...
-                            throw new IOException("Unrecognized Proc-Type: " + line.substring(11));
+                            throw new FormatException("Unrecognized Proc-Type: " + line.substring(11));
                         }
                     } else if (line.startsWith("DEK-Info: ")) {
                         int ptr = line.indexOf(",");
                         if (ptr == -1) {
-                            throw new IOException("Unrecognized DEK-Info: " + line.substring(10));
+                            throw new FormatException("Unrecognized DEK-Info: " + line.substring(10));
                         } else {
                             String algorithm = line.substring(10,ptr);
```
hierynomus: That one seems like a false positive, i.e. this is a local variable not a field.

joval:
I transplanted the test... I need to put the test into my build workflow so this doesn't happen again.

Do you really want me to make the new test class less readable per Codacy?

hierynomus:
Actually the question would be, why are they static final fields if they're
needed by one test? I do agree with the codestyle here, move them to the
top, or make them local variables.

2016-08-19 22:11 GMT+02:00 JovalCM <notifications@github.com>:

joval:
I just copied from the OpenSSHKeyFileTest.  The id_dsa file (used by that test) is an encrypted PKCS5 key file, and its test checks the exact use-case that was broken.  Constants are static final fields in all the keyfile tests.

Issue #264, opened 2016-08-17
Per-client logging


joval:
Would you be amenable to my adding a setLog(org.slf4j.Logger) method to net.schmizz.sshj.SSHClient?  (Meaning the log member variable would no longer be final).

When working with hundreds of connections in parallel, I think it's extremely useful to be able to separate out the log events for each one.

By default (i.e., if the setLog method is not called), I'd leave things as-is.

hierynomus:
The problem is that each of the objects underneath the SSHClient has its own logger, so I don't think this will add much. You would normally do this using the MDC (message diagnostic context).
Or you can add the Thread to the log pattern, making it easier to filter out the per-thread log messages.

So I don't think having a `setLog(...)` method is very useful, as you would only change the logger of `SSHClient`.

joval:
Well, the underlying implementation would be more complex, as I would want to set the Loggers for all the classes downstream from the SSHClient class on a per-client basis as well.

Using MDC, I would have to inject context information into every sub-logger as well -- which is essentially the same problem, except much harder to use from logging frameworks like JUL.

I have never found class-oriented logging to be particularly useful in the context of a connection-oriented system, considering that logging frameworks can generally tell you the class origin of each log message.

If I can leave the default logging behavior as-is, would you be open to adding such a feature?

hierynomus:
Well actually I'm not in favour of this. Needing to propagate the same logger throughout the codebase feels like an absolute anti-pattern to me. Do you have an example of _any_ framework/library which does this?

The MDC is actually really the correct way of doing this. The MDC is shared in the thread as a ThreadLocal. So there is no need to propagate this to every Logger. From your own app you can just do:

```
try(MDC.MDCCloseable closeable = MDC.putCloseable("ssh-host", <ip>)) {
     sshClient.connect(<ip>);
     sshClient.authPassword(...);
    ...
}
```

Another option, and one I'm also not fond of, is introducing a wrapper class around the LoggerFactory, which is then used to lookup the loggers. But again this means that all loggers become instance variables, which is not something I'm fond of.

joval:
I was a little confused because every logger in sshj is already an instance variable (only final).  If you'd really wanted to enforce the current pattern they should be static final.

I haven't studied the threading behavior of sshj.  Are there not potentially multiple threads per SSH connection?  Or are all the loggers assigned in the current thread in the SSHClient.connect method?

Overriding the factory will not work, because org.slf4j.LoggerFactory is declared final.

[...]

PR #267, opened 2016-08-24
Logging factory

solind:
Here is a separate PR for the LoggingFactory injection we discussed in issue #264.


solind:
I don't understand why the license check failed, I copied the Copyright from another source file in the project.

solind:
Regarding the remaining nits from codacy, as you can see this is a pretty large bunch of changes.  It's complaining about member variable declarations that are essentially unmoved from the pre-refactored code base, and I had no intention of re-factoring the classes to meet the new style guidelines.  I hope that doesn't pose a significant issue.

src/main/java/com/hierynomus/sshj/transport/IdentificationStringParser.java
```diff
@@ -17,20 +17,21 @@
...
 public class IdentificationStringParser {
-    private static final Logger logger = LoggerFactory.getLogger(IdentificationStringParser.class);
+    private final Logger log;
     private final Buffer.PlainBuffer buffer;
 
     private byte[] EXPECTED_START_BYTES = new byte[] {'S', 'S', 'H', '-'};
 
-    public IdentificationStringParser(Buffer.PlainBuffer buffer) {
+    public IdentificationStringParser(LoggerFactory loggerFactory, Buffer.PlainBuffer buffer) {
+	this.log = loggerFactory.getLogger(IdentificationStringParser.class);
```
hierynomus: indentation is wrong

src/main/java/net/schmizz/concurrent/Promise.java
```diff
@@ -60,10 +60,11 @@ public Promise(String name, ExceptionChainer<T> chainer) {
      * @param chainer {@link ExceptionChainer} that will be used for chaining exceptions
      * @param lock    lock to use
      */
-    public Promise(String name, ExceptionChainer<T> chainer, ReentrantLock lock) {
+    public Promise(String name, ExceptionChainer<T> chainer, ReentrantLock lock, LoggerFactory loggerFactory) {
         this.name = name;
         this.chainer = chainer;
         this.lock = lock == null ? new ReentrantLock() : lock;
+	this.log = loggerFactory.getLogger(getClass());
```
hierynomus: indentation

src/main/java/com/hierynomus/sshj/transport/IdentificationStringParser.java
solind: I fixed the indentation (sorry I use tabs for 8 spaces reflexively).  What’s up with the “license” issue?

hierynomus: If you add new classes, easiest is to run `gradle licenseFormat`. This will fix the license headers.

src/main/java/net/schmizz/sshj/common/KeyType.java
```diff
@@ -174,15 +174,15 @@ protected boolean isMyType(Key key) {
     },
 
     ED25519("ssh-ed25519") {
-        private final Logger logger = LoggerFactory.getLogger(KeyType.class);
+        private final Logger log = LoggerFactory.getLogger(KeyType.class);
```
hierynomus: Isn't this using the wrong `LoggerFactory` now?

src/main/java/net/schmizz/sshj/sftp/RemoteDirectory.java
```diff
@@ -25,7 +25,7 @@
 public class RemoteDirectory
         extends RemoteResource {
 
-    public RemoteDirectory(Requester requester, String path, byte[] handle) {
+    public RemoteDirectory(SFTPEngine requester, String path, byte[] handle) {
```
hierynomus: Why?

src/main/java/net/schmizz/sshj/sftp/RemoteFile.java
```diff
@@ -30,7 +30,7 @@
 public class RemoteFile
         extends RemoteResource {
 
-    public RemoteFile(Requester requester, String path, byte[] handle) {
+    public RemoteFile(SFTPEngine requester, String path, byte[] handle) {
```
hierynomus: Why?

src/main/java/net/schmizz/sshj/DefaultConfig.java
```diff
@@ -84,6 +85,30 @@ public DefaultConfig() {
         setKeepAliveProvider(KeepAliveProvider.HEARTBEAT);
     }
 
+    /**
+     * Default SLF4J-based implementation of the SSHJ LoggerFactory.
+     */
+    public static class DefaultLoggerFactory implements LoggerFactory {
```
hierynomus: Is this one still needed?

src/main/java/net/schmizz/sshj/sftp/RemoteDirectory.java
solind: I added a protected getLoggingFactory to SFTPEngine.  If you’d prefer I could add the method to the Requester interface.  Since SFTPEngine, RemoteDirectory and RemoteFile are all in the same package, it seemed to me that treating the Requester like an SFTPEngine in the constructors would be harmless.

src/main/java/net/schmizz/sshj/common/KeyType.java
solind: There are a few places — like these static initializers — where it’s not possible to inject a LoggerFactory.  I could use the new LoggerFactory.DEFAULT instead, but instead I just left them as-is, as the result is identical.

src/main/java/net/schmizz/sshj/DefaultConfig.java
solind: Oops, no, I forgot to remove that.

solind:
The PacketReaderTest fails with a NullPointerException because a mock STFPEngine's getLoggerFactory method returns null.

It seems silly to do a null check for a final variable... it also seems silly to vastly complicate the packet reader test by creating a real SFTPEngine.

I'll see if I can use Mockito to return a LoggerFactory.DEFAULT.

solind:
Hi Jeroen, any thoughts on this PR?  I have a few additional modifications that are waiting in the wings (not logging-related).

hierynomus:
I'll have a look after the weekend. I've quickly looked through it, but want a more in-depth look before I merge it.

solind:
Hey Jeroen, any code review feedback yet?  The feature is working well in my integration with our product, I'm eager to move forward.

Issue #234, opened 2016-02-10
sshj can't run with java 6


sblond:
Hi,

I get `java.lang.UnsupportedClassVersionError: net/i2p/crypto/eddsa/spec/EdDSAParameterSpec : Unsupported major.minor version 51.0` when using sshj 0.15.0 with java 6. Usually this error means that a class has been compiled with java 7 compatibility but run with an older java.

This may be related to the fix of #171 that adds a new dependency to net.vrallev.ecc:ecc-25519-java:1.0.1. This dependency seems to require the use of java 7 https://github.com/vRallev/ECC-25519/blob/master/ECC-25519-Java/build.gradle.

Does it mean that java 6 compatibility is dropped?

hierynomus:
Hi Stephane,

Thanks for pointing that out. Hadn't actually noticed that!
I am somewhat inclined to then maybe indeed drop java6 support for new versions. Or to make ecc-25519 an optional dependency, and checking in the code whether it can be added or not.
What would be the preference?

dkocher:
+1 to drop Java 6 support.

sblond:
I have to deal with java 6, so
+1 to keep java 6 support

hierynomus:
I cannot easily make ecc-25519 optional, as it is used directly in the `KeyType` enum, so I cannot optionally load it. So I am dropping java6 support, sorry.

hierynomus:
@sblond We've been able to add Java6 support back to the mix.

sblond:
Hi,
I upgraded to 0.18.0 and it worked fine.
Thanks!

PR #269, opened 2016-09-02
Compatibility features

solind:
These are the remaining changes I had to add to sshj for compatibility with Joval's jSAF provider code.  They're all relatively minor so I hope it makes sense to group them together:
1) Re-wrote try-with-resources and inferred generics constructors (no functional changes) so that the project can be built using Java 6.
2) Added a method to the Channel interface to make it possible to determine whether an EOF has been received.
3) Added some missing SFTP response codes.
5) Fixed a typo in an SFTP enum.

### DIFF
--- a/src/main/java/net/schmizz/sshj/connection/channel/AbstractChannel.java
+++ b/src/main/java/net/schmizz/sshj/connection/channel/AbstractChannel.java
@@ -52,6 +52,8 @@ public abstract class AbstractChannel
     /** Remote recipient ID */
     private int recipient;
 
+    private boolean eof = false;
+
     private final Queue<Event<ConnectionException>> chanReqResponseEvents = new LinkedList<Event<ConnectionException>>();
 
     /* The lock used by to create the open & close events */
@@ -189,6 +191,11 @@ public abstract class AbstractChannel
                 gotUnknown(msg, buf);
 
         }
+    }
+
+    @Override
+    public boolean isEOF() {
+	return eof;
     }
 
     @Override
@@ -394,6 +401,7 @@ public abstract class AbstractChannel
     /** Called when EOF has been received. Subclasses can override but must call super. */
     protected void eofInputStreams() {
         in.eof();
+	eof = true;
     }
 
     @Override
--- a/src/main/java/net/schmizz/sshj/connection/channel/AbstractChannel.java
+++ b/src/main/java/net/schmizz/sshj/connection/channel/AbstractChannel.java
@@ -195,12 +195,12 @@ public abstract class AbstractChannel
 
     @Override
     public boolean isEOF() {
-	return eof;
+        return eof;
     }
 
     @Override
     public LoggerFactory getLoggerFactory() {
-	return loggerFactory;
+        return loggerFactory;
     }
 
     private void gotClose()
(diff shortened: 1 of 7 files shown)


hierynomus:
Hi David, unfortunately Java6 compatibility is a no-go, as per #234, the ed-25519 library we depend on is java7 only.

solind:
Interesting.  I actually compiled the ed25519-java project using Java 6, made a Java 6 build of sshj and it worked fine.  (We don't allow build frameworks to pull down dependencies for production builds, we compile everything from source).

hierynomus:
Hi David,

Unfortunately the ed25519-java project is released only with a Java7 build. So I also only support java7. Also because Java6 is EOL, and Java7 is also officially EOL-ed by Oracle, see: http://www.oracle.com/technetwork/java/eol-135779.html

So for me, adding back in Java6 support is a step backwards, also because my CI builds run with java7+, and do not test for java6 compatibility anymore. So there is absolutely no guarantee that these changes will be kept in.

solind:
I know, they're dinosaurs, but I think we have customers who are still using it.  I see it more on AIX and old versions of Linux using OpenJDK (not Oracle's VM).

How about this, what if I take responsibility for any Java 6 compatibility work?  Feel free to break Java 6 compatibility and if it bothers me, I'll fix it?  You won't have to worry about it, and the project can officially still require Java 7.

(Obviously though, you'll have to leave me the option of eventually giving up on trying to maintain compatibility.)

hierynomus:
Hi David,

I had another quick look at the code. Essentially the `ed25519-java` library indeed is released with Java6, however the `VRallev/ECC-25519` library is released with Java7. The latter one bundles the former one.

Looking through the code. I just use the `djb.Curve25519` class from that library. Which is a Public Domain class.

What I would propose is that we can add back Java6 compatibility by including the `djb.Curve25519` class in the codebase, and depending on the `net.i2p.crypto:ed25519-java` library. I'll fix this momentarily and then with your changes I'll re-add java6 to the compilation targets on the TravisCI server.

Sound like a plan?

solind:
Sounds awesome!  Thanks!

src/main/java/net/schmizz/sshj/connection/channel/AbstractChannel.java
```diff
@@ -394,6 +401,7 @@ private void gotEOF()
     /** Called when EOF has been received. Subclasses can override but must call super. */
     protected void eofInputStreams() {
         in.eof();
+	eof = true;
```
hierynomus: Somehow your indenting is off sometimes ;)

hierynomus:
Can you merge in master so that the checks are green ;)

solind:
I'm sorry man.  I actually write code in vi (speaking of dinosaurs), and I normally use 1 tab for 8 spaces.

hierynomus:
I think the right dinosaur settings are:

```
set autoindent
set smartindent
set expandtab
set shiftwidth=4
set tabstop=4
``` 

😉

hierynomus:
Sorry clicked wrong button there... If you can get the branch up-to-date then we're good to go!

solind:
I think I just did. :)

PR #279, opened 2016-10-28
Support for OpenSSH new key file format 

hierynomus:
The new format is used for ed-25519 keys.

Only support for non-encrypted keys currently.


hierynomus:
@dkocher Can you test it out?

dkocher:
@hierynomus Will do today!

hierynomus:
It's a bit later then I wanted, but it took some experimenting ;)

src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
```diff
@@ -0,0 +1,160 @@
...
+            throw new IOException("We don't support having more than 1 key in the file (yet).");
+        }
+        PublicKey publicKey = readPublicKey(new PlainBuffer(keyBuffer.readBytes())); // string publickey1
+        PlainBuffer privateKeyBuffer = new PlainBuffer(keyBuffer.readBytes()); // string (possibly) encrypted, padded list of private keys
+        if ("none".equals(cipherName)) {
+            logger.debug("Reading unencrypted keypair");
+            return readUnencrypted(privateKeyBuffer, publicKey);
+        } else {
+            logger.info("Keypair is encrypted with: " + cipherName + ", " + kdfName + ", " + kdfOptions);
+            throw new IOException("Cannot read encrypted keypair with " + cipherName + " yet..");
```
dkocher: What additional work is required to support encrypted keys?

dkocher:
I have tested this with a private key of mine and it works as advertised.

src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
hierynomus: I have to figure out how the 'altered' bcrypt implementation works... i.e. seems non-trivial at the moment.

hierynomus:
Yay! If you can think of a better class name I'll rename it. Else I'll leave that for a moment.

Going through this code again, it could use a lot of cleanup. I want to split off the keyfile format from the keytype stored in there. Will do that separately though.

PR #291, opened 2016-12-21
Two minor changes

solind:
1) We shouldn't assume that the current thread context ClassLoader has access to the sshj.properties file, but it's safe to assume that the DefaultConfig class's ClassLoader does.

2) When creating an SSHRuntimeException based on a Throwable, I believe the SSHRuntimeException should adopt the message of the Throwable cause (instead of always using null).

WDYT?


hierynomus:
Agreed on both accounts :) Thanks!

Issue #255, opened 2016-07-17
OSGI dependencies are broken


15knots:
The manifest header declares a dependency to package `net.i2p.crypto.eddsa.math`, but bundle `net.i2p.crypto.eddsa` does not export that package. (The pom of net.i2p.crypto.eddsa explicitly excludes the package from the exported package list.)

This results in a tycho error when building:
Missing requirement: com.hierynomus.sshj 0.17.2 requires 'package net.i2p.crypto.eddsa.math 0.0.0' but it could not be found

hierynomus:
As I'm not an OSGI whiz, what would be the best solution for this? Isn't this something that needs to be fixed in the `net.i2p.crypto.eddsa` pom?

15knots:
src/main/java/net/schmizz/sshj/common/KeyType.java imports the package and uses types from it.

So it needs to be fixed in the net.i2p.crypto.eddsa pom.

15knots:
Thank you for the quick response!

15knots:
Please publish a new release containing this fix. Thank You!

hierynomus:
Done, version 0.18.0 is downloadable from Maven.

adagios:
Unfortunately it's still broken. The build.gradle file also needs to be corrected. I've created a pull request.

15knots:
I am currently working on a project that tests how sshj behaves in an OSGI environment.
Things I found so far:

1. `Missing requirement: com.hierynomus.sshj 0.19.0 requires 'package net.i2p.crypto.eddsa.math 0.0.0' but it could not be found.` This is because sshj imports package net.i2p.crypto.eddsa.math, which is not exported (possibly a bug in gradle bundler which adds the import?)

1. `ClassNotFoundEx: org.bouncycastle.jce.provider.BouncyCastleProvider in net.schmizz.sshj.common.SecurityUtils.BouncyCastleRegistration.run()` due to a missing dependency in manifest.mf.

1.  Shouldn´ t sshj export packages `com.hierynomus.sshj.*`, too?

I fixed all of the above by changing `build.gradle`this way:
```
jar {
    manifest {
      // please see http://bnd.bndtools.org/chapters/390-wrapping.html
        instruction "Bundle-Description", "SSHv2 library for Java"
        instruction "Bundle-License", "http://www.apache.org/licenses/LICENSE-2.0.txt"
        instruction "Import-Package", \
[...]

PR #295, opened 2017-01-12
converts bouncycastle OSGi dep from Import-Bundle to Import-Package

adagios:
Another attempt at resolving #255 

Import-Bundle is generally discouraged by OSGi best practices.

See http://stackoverflow.com/questions/1865819/when-should-i-use-import-package-and-when-should-i-use-require-bundle for more info.


hierynomus:
You've forgotten `com.hierynomus.sshj` ;)

@15knots does it work for you too?

adagios:
I wrote on the original bug:

> With regards to exporting `com.hierynomus.sshj` packages, is it something that clients should use directly? I looked and they seemed like internal classes, so I left them out of the `Export-Package`.

should I add it?

hierynomus:
Yes, you'd need them if you want to customize the config to use specific algorithms only

adagios:
done!

hierynomus:
OSGI n00b question: Should we then not also add a negative `Import-Package` rule for `com.hierynomus.sshj`?

adagios:
All exported packages are normally also imported to prevent multiple class spaces.

See http://blog.osgi.org/2007/04/importance-of-exporting-nd-importing.html for a thorough explanation .

Issue #277, opened 2016-10-24
Charset support


ISQ-GTT:
We would like to use sshj to communicate with various machines which are configured to use ISO 8859-1. 
When using sshj, received byte streams which are encoded in ISO 8859-1 get decoded using UTF-8 by default resulting in encoding problems with Umlaute and such.
There's no possibiliy for us to change the encoding on server-side or extend sshj-code to support that charset.

All encoding problems seem to break down to the net.schmizz.sshj.common.Buffer classes readString()- and putString()-methods where UTF-8 is used hard-coded.

It would be great to support arbitrary charsets like other SSH-frameworks (e.g. jsch) or SSH-tools (e.g. PuTTY).

hierynomus:
Do you have an example, and a more specific place on where this is happening? File copies, shell, exec's...

ISQ-GTT:
It applies to exec's, shell-operations and file copies.

For example:
- I created a remote file "TestFileÄÖÜ.txt" using SFTP, SCP, Exec and Shell each. 
In WinSCP it gets displayed as "TestFileÃ„Ã–Ãœ.txt". In PuTTY it gets displayed in 2 lines like "TestFileÃ\nÃÃ.txt".
- When listing files in sshj using "ls" or SFTPClient.ls(), a file "TestFileÄÖÜ.txt" that was remotely created using PuTTY or WinSCP gets displayed as "TestFile???.txt".
Calling getBytes() returns "84, 101, 115, 116, 70, 105, 108, 101, -17, -65, -67, -17, -65, -67, -17, -65, -67, 46, 116, 120, 116". Each '?' is encoded as "-17, -65, -67".
On the other hand, all files created with sshj will be displayed correctly.
- Existing files can't be found. Executing "cat TestFileÄÖÜ.txt" results in "cat: 0652-050 Cannot open TestFileÄÖÜ.txt." and only works if the file was created using sshj, e.g. when "TestFileÃ„Ã–Ãœ.txt" exists. Same for SFTP/SCP downloads

hierynomus:
Can you write a (unit) test to easily simulate this behaviour? I'd be more than happy to fix this.

ISQ-GTT:
When creating the file with sshj, the UTF-8 filename is remotely decoded with the local charset resulting in an odd filename.
Displaying that filename on sshj-side reverses that process so it will look normal again to the sshj-client.
Basically sshj-created files act as expected while using sshj. 

[...]

PR #305, opened 2017-03-08
Added charset support, centralized UTF-8 usage

ISQ-GTT:
Here's the fix as promised in Issue #277. Feel free to make changes.

### DIFF
--- a/src/main/java/net/schmizz/sshj/SSHClient.java
+++ b/src/main/java/net/schmizz/sshj/SSHClient.java
@@ -449,9 +449,8 @@ public class SSHClient
      *
      * @return remote character set or {@code null} for default
      */
-    public Charset getRemoteCharset()
-    {
-    	return remoteCharset;
+    public Charset getRemoteCharset() {
+        return remoteCharset;
     }
 
     /** @return a {@link RemotePortForwarder} that allows requesting remote forwarding over this connection. */
@@ -722,15 +721,14 @@ public class SSHClient
         doKex();
     }
 
-	/**
-	 * Sets the character set used to communicate with the remote machine for certain strings (like paths)
-	 *
-	 * @param remoteCharset
-	 *        remote character set or {@code null} for default
-	 */
-    public void setRemoteCharset(Charset remoteCharset)
-    {
-    	this.remoteCharset = remoteCharset;
+    /**
+     * Sets the character set used to communicate with the remote machine for certain strings (like paths)
+     *
+     * @param remoteCharset
+     *        remote character set or {@code null} for default
+     */
+    public void setRemoteCharset(Charset remoteCharset) {
+        this.remoteCharset = remoteCharset;
     }
 
     @Override
--- a/src/main/java/net/schmizz/sshj/SSHClient.java
+++ b/src/main/java/net/schmizz/sshj/SSHClient.java
@@ -61,6 +61,7 @@ import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.net.ServerSocket;
+import java.nio.charset.Charset;
 import java.security.KeyPair;
 import java.security.PublicKey;
 import java.util.*;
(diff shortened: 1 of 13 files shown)


src/main/java/net/schmizz/sshj/SSHClient.java
```diff
@@ -440,6 +444,16 @@ public Connection getConnection() {
         return conn;
     }
 
+    /**
+     * Returns the character set used to communicate with the remote machine for certain strings (like paths).
+     *
+     * @return remote character set or {@code null} for default
+     */
+    public Charset getRemoteCharset()
+    {
```
hierynomus: Please use the same formatting as the rest of the code, ie. open-curly on end-of-line

src/main/java/net/schmizz/sshj/connection/channel/direct/SessionChannel.java
```diff
@@ -93,7 +98,7 @@ public Command exec(String command)
             throws ConnectionException, TransportException {
         checkReuse();
         log.debug("Will request to exec `{}`", command);
-        sendChannelRequest("exec", true, new Buffer.PlainBuffer().putString(command))
+        sendChannelRequest("exec", true, new Buffer.PlainBuffer().putString(command.getBytes(getRemoteCharset())))
```
hierynomus: The main question I have when reading this, should _every_ string be written using the remote charset? If so, it is nicer to make it a property of the Buffer, rather than converting it everywhere we call `putString`.

hierynomus: Or make a `putString(String s, Charset cs)` method, which makes the rest of the code at least cleaner and more concise.

ISQ-GTT: The remote side seems to decode the command using its own charset instead of utf-8. Commands containing filenames with special characters won't work properly using utf-8 on non-utf-8 remote machines.
I'll add the `putString(String s, Charset cs)` method. There will still be a lot of `getRemoteCharset()` calls in SFTPEngine though which could also be replaced by a property.

src/main/java/net/schmizz/sshj/SSHClient.java
```diff
@@ -128,6 +129,9 @@
 
     private final List<LocalPortForwarder> forwarders = new ArrayList<LocalPortForwarder>();
 
+    /** character set of the remote machine */
+    protected Charset remoteCharset;
```
hierynomus: Assign the default `IOUtils.UTF_8` here.

src/main/java/net/schmizz/sshj/SSHClient.java
```diff
@@ -708,12 +721,22 @@ public void rekey()
...
     }
 
+    /**
+     * Sets the character set used to communicate with the remote machine for certain strings (like paths)
+     *
+     * @param remoteCharset
+     *        remote character set or {@code null} for default
+     */
+    public void setRemoteCharset(Charset remoteCharset) {
+        this.remoteCharset = remoteCharset;
```
hierynomus: Ensure that you cannot set `null`

src/main/java/net/schmizz/sshj/SSHClient.java
```diff
@@ -708,12 +721,22 @@ public void rekey()
...
+    }
+
     @Override
     public Session startSession()
             throws ConnectionException, TransportException {
         checkConnected();
         checkAuthenticated();
-        final SessionChannel sess = new SessionChannel(conn);
+        final SessionChannel sess = (remoteCharset != null)? new SessionChannel(conn, remoteCharset) : new SessionChannel(conn);
```
hierynomus: No need to do the if in the case that we default to UTF_8 here already.

hierynomus:
THanks, merged :)

Issue #153, opened 2014-10-16
ssh certificate authentication fails


WellRedPandit:
sshj fails where openssh succeeds without a password using two files `private-key` and `private-key-cert.pub`. What am I coding wrongly with sshj?

(if sshj does not support certificate authentication, am willing to contribute, if pointed in the right direction...)
### Test

``` scala
object Sshj {
  def ping() = {
    val ssh = new SSHClient
    ssh.loadKnownHosts()
    ssh.connect("555.16.13.210", 26)
    try {
      ssh.authPublickey("root","~/.ssh/private-key")
      val session = ssh.startSession()
      try {
        val cmd = session.exec("ping -c 1 google.com")
        println(IOUtils.readFully(cmd.getInputStream()).toString())
        cmd.join(5, TimeUnit.SECONDS)
        println("\n** exit status: " + cmd.getExitStatus())
      } finally {
        session.close()
      }
    } finally {
      ssh.disconnect()
    }
  }
}
```
### Scala REPL

```
scala> Sshj.ping
[INFO ] [15:05:15.179] [run-main-4] [n.schmizz.sshj.common.SecurityUtils] BouncyCastle already registered as a JCE provider
[WARN ] [15:05:15.195] [run-main-4] [net.schmizz.sshj.DefaultConfig] Disabling high-strength ciphers: cipher strengths apparently limited by JCE policy
[INFO ] [15:05:15.232] [run-main-4] [n.s.sshj.transport.TransportImpl] Client identity string: SSH-2.0-SSHJ_0_9_0
[INFO ] [15:05:15.236] [run-main-4] [n.s.sshj.transport.TransportImpl] Server identity string: SSH-2.0-OpenSSH_5.9
[DEBUG] [15:05:15.237] [run-main-4] [net.schmizz.concurrent.Promise] Setting <<kex done>> to `null`
[DEBUG] [15:05:15.237] [run-main-4] [n.s.sshj.transport.KeyExchanger] Sending SSH_MSG_KEXINIT
[DEBUG] [15:05:15.248] [run-main-4] [net.schmizz.concurrent.Promise] Setting <<kexinit sent>> to `SOME`
[DEBUG] [15:05:15.249] [run-main-4] [net.schmizz.concurrent.Promise] Awaiting <<kex done>>
[DEBUG] [15:05:15.249] [reader] [n.s.sshj.transport.KeyExchanger] Received SSH_MSG_KEXINIT
[...]

PR #319, opened 2017-04-12
Add support for authentication with DSA & RSA user certificates (#153)

chqr:
First pass at support for RSA & DSA certificate authentication. Thanks for maintaining this awesome library!

Updates:
- KeyType.java - add support for two certificate key types
    ssh-rsa-cert-v01@openssh.com
    ssh-dsa-cert-v01@openssh.com

- Buffer.java - allow uint64s that overflow Long.MAX_VALUE, otherwise
  we break on certificates with serial numbers greater Long.MAX_VALUE

- OpenSSHKeyFile, KeyProviderUtil - prefer public key files that end
  "-cert.pub" if they exist

Added new class Certificate, which represents certificate key

Reference:

https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/PROTOCOL.certkeys?annotate=HEAD


hierynomus:
Hi @chqr,

Thanks for the PR. Looks pretty good already.
One of my main concerns is the change in Buffer currently. I think rather than doing that, we should be using `readMPInt` here, which allows 'arbitrary big values', and thus doesn't suffer from the overflow that UInt64 is prone to. Judging from the code that wouldn't be causing any problems at first sight. Do you agree?

chqr:
Hi @hierynomus,

Thanks for the fast response! Looks like mpints are encoded differently on the wire from uint64s, so they can't be used interchangeably. But I could add new methods readUInt64AsBigInteger() and putUInt64(BigInteger uint64), and leave the size checks in place for the long methods?

Will look at the codacy issues.

src/main/java/net/schmizz/sshj/common/KeyType.java
```diff
@@ -266,4 +322,153 @@ public static KeyType fromString(String sType) {
...
+                type = buf.readUInt32();
+                id = buf.readString();
+                validPrincipals = unpackList(buf.readStringAsBytes());
+                validAfter = dateFromEpoch(buf.readUInt64());
+                validBefore = dateFromEpoch(buf.readUInt64());
+                critOptions = unpackMap(buf.readStringAsBytes());
+                extensions = unpackMap(buf.readStringAsBytes());
+                buf.readString(); // reserved
+                signatureKey = buf.readStringAsBytes();
+                signature = buf.readStringAsBytes();
```
hierynomus: Why not `readBytes()`

src/main/java/net/schmizz/sshj/userauth/keyprovider/OpenSSHKeyFile.java
```diff
@@ -57,10 +57,14 @@ public PublicKey getPublic()
 
     @Override
     public void init(File location) {
-        final File f = new File(location + ".pub");
-        if (f.exists())
+        // try cert key location first
+        File pubKey = new File(location + "-cert.pub");
+        if (! pubKey.exists()) {
```
hierynomus: formatting, there should be no space between `!` and statement.

src/main/java/net/schmizz/sshj/common/KeyType.java
```diff
@@ -266,4 +322,153 @@ public static KeyType fromString(String sType) {
...
+            buf.putUInt64(certificate.getSerial())
+                .putUInt32(certificate.getType())
+                .putString(certificate.getId())
+                .putString(packList(certificate.getValidPrincipals()))
+                .putUInt64(epochFromDate(certificate.getValidAfter()))
+                .putUInt64(epochFromDate(certificate.getValidBefore()))
+                .putString(packMap(certificate.getCritOptions()))
+                .putString(packMap(certificate.getExtensions()))
+                .putString("") // reserved
+                .putString(certificate.getSignatureKey())
```
hierynomus: `putBytes` (also for other places). Esp. signatures and keys are not strings but rather `byte[]`.
Though the implementation in the buffer doesn't differ, the semantics are different.

hierynomus:
Had a bit more in-depth look this morning, few questions/issues left.

hierynomus:
Thanks for the great contribution!

Issue #306, opened 2017-03-08
Defect introduced in code cleanup commit?


solind:
I'm seeing the following exception sending really long strings down a ChannelOutputString, which I don't see when using a previous build:

```
java.lang.ArrayIndexOutOfBoundsException: null
        at net.schmizz.sshj.common.Buffer.putRawBytes(Buffer.java:266)
        at net.schmizz.sshj.connection.channel.ChannelOutputStream$DataBuffer.write(ChannelOutputStream.java:64)
        at net.schmizz.sshj.connection.channel.ChannelOutputStream.write(ChannelOutputStream.java:141)
```

I looked cursorily and didn't see a problem, but I suspect this might have been broken in commit https://github.com/hierynomus/sshj/commit/636f8968503ba9b80cea077d17a65eb331a3be8e.

hierynomus:
Can you create a unit test that simulates this?

solind:
I'm super-busy at the moment, but I will give it a shot if you'll point me to an existing unit test that attempts to execute any command over an exec or shell channel.

If you can't, it'll just take me longer to get around to it.

hierynomus:
Hi David, else do you have some logging and string sizing info for me?

Op 9 mrt. 2017 6:37 p.m. schreef "David Solin" <notifications@github.com>:

solind:
Hi Jeroen, sorry I never managed to get that unit test working.  I just tried this update and it appears fixed to me.  Thanks!

hierynomus:
No problem, the keen eyes of @dkocher found the issue ;)

PR #322, opened 2017-04-25
Fix regression from 40f956b. Invalid length parameter.

dkocher:


hierynomus:
Thanks @dkocher! I think this fixes #306. Good find totally missed that

dkocher:
Would be nice if there is some test coverage that would catch this.

hierynomus:
Agreed, do you have time to add a test? :)

hierynomus:
I'll merge this for now and release a new release.

PR #336, opened 2017-06-25
Ecdsa sha2++

Igerly:
Hey there!

I've added support for ecdsa-sha2-nistp384 and ecdsa-sha2-nistp521 in this PR. The *521 one also uncovered a problem in the way you've been asn1-encoding stuff, so I've switched that from "manual" to a more "official" way.


hierynomus:
Hi, Thanks for the PR! :) Great work. Could you add some tests with the different keytypes, especially for the bug found in the ASN.1 encoding? THen it is complete and ready to be merged.

Igerly:
Sure, just a couple of questions :)

- With the keytypes, do you mean the ones alike to the yesterday's failed one - computing the fingerprints? 
- I haven't been able to find any tests which would actually connect to a kind of ssh server - are there any (I was testing these against a local Bitvise server, but, perhaps, openssh or something could also be used)?

hierynomus:
Yes, please add at least tests for the failures you found, that would prevent regressions.

There are some tests that connect to an Apache Mina SSH server, which is booted up in memory. But I would also welcome some (Dockerized) OpenSSH tests, that would be awesome.

Igerly:
I've added the unit-tests for the cases. 
I've also started to look into the dockerized integration version, but will add these in another PR, if they turn out working :)

hierynomus:
Looks great, thanks! I'll do some tests tomorrow and merge it afterwards. 

Looking forward to your next pull request. 

_Sent from my Samsung SM-G950F using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #338, opened 2017-07-06
Add ConsolePasswordFinder to read from Console

matthew-dailey:
* I didn't find a `PasswordFinder` to prompt a user for their password,
so this adds that implementation

### DIFF
--- a/src/main/java/net/schmizz/sshj/userauth/password/ConsolePasswordFinder.java
+++ b/src/main/java/net/schmizz/sshj/userauth/password/ConsolePasswordFinder.java
@@ -21,17 +21,24 @@ import java.util.IllegalFormatException;
 /** A PasswordFinder that reads a password from a console */
 public class ConsolePasswordFinder implements PasswordFinder {
 
+    public static final String DEFAULT_FORMAT = "Enter passphrase for %s:";
+
     private final Console console;
     private final String promptFormat;
     private final int maxTries;
 
     private int numTries;
 
-    public static ConsolePasswordFinderBuilder builder() {
-        return new ConsolePasswordFinderBuilder();
+    public ConsolePasswordFinder() {
+        this(System.console());
+    }
+
+    public ConsolePasswordFinder(Console console) {
+        this(console, DEFAULT_FORMAT, 3);
     }
 
     public ConsolePasswordFinder(Console console, String promptFormat, int maxTries) {
+        checkFormatString(promptFormat);
         this.console = console;
         this.promptFormat = promptFormat;
         this.maxTries = maxTries;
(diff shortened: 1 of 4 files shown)


src/main/java/net/schmizz/sshj/userauth/password/ConsolePasswordFinder.java
```diff
@@ -0,0 +1,52 @@
...
+        this.console = console;
+    }
+
+    @Override
+    public char[] reqPassword(Resource<?> resource) {
+        if (console == null) {
+            // the request cannot be serviced
+            return null;
+        }
+        return console.readPassword("Enter passphrase for %s:", resource.toString());
```
hierynomus: Let's make the String a parameter to the class, in that way users can customize the message the user sees.

Also let's add a max number of retries as a parameter.

src/test/java/net/schmizz/sshj/userauth/password/TestConsolePasswordFinder.java
```diff
@@ -0,0 +1,36 @@
...
+package net.schmizz.sshj.userauth.password;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class TestConsolePasswordFinder {
+
+    /*
+     * Note that Mockito 1.9 cannot mock Console because it is a final class,
```
hierynomus: I'm fine with upgrading to a newer Mockito version :)

src/test/java/net/schmizz/sshj/userauth/password/TestConsolePasswordFinder.java
```diff
@@ -0,0 +1,36 @@
...
+     * so there are no other tests.
+     */
+
+    @Test
+    public void testReqPasswordNullConsole() {
+        char[] password = new ConsolePasswordFinder(null)
+            .reqPassword(Mockito.mock(Resource.class));
```
hierynomus: Any reason to mocking the Resource here? You don't verify it's not being called upon...

hierynomus:
Nice improvement, few comments before I'll merge it.

matthew-dailey:
Thanks for the feedback!  I should be able to make those changes some time today, and I'll work with upgrading Mockito to see if any other code breaks.

hierynomus:
Ok! I look forward to the updates :)

matthew-dailey:
This is ready for another look whenever you're free.
I ended up adding a lot more code than I thought, but I figured the Builder would make the class easier to use.
I also upgraded Mockito to the absolute latest, which surprisingly worked :)

src/main/java/net/schmizz/sshj/userauth/password/ConsolePasswordFinder.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private int numTries;
+
+    public static ConsolePasswordFinderBuilder builder() {
+        return new ConsolePasswordFinderBuilder();
+    }
+
+    public ConsolePasswordFinder(Console console, String promptFormat, int maxTries) {
+        this.console = console;
+        this.promptFormat = promptFormat;
```
hierynomus: You're not checking the format of the prompt here, but the constructor is public.

I don't think we actually need the builder, it's just 3 parameters. I'm fine with using just a constructor.

In any case we need to do the check here also.

matthew-dailey: Good catch!  OK, I'll take out the builder (good to have less code to maintain) and put the check here

PR #357, opened 2017-09-05
Disambiguated signature initialization

charlesrgould:
Initially (hah), the signature API was slightly unclear in that if you called `Signature#initialize(PublicKey, PrivateKey)` with non-null `publicKey` and `privateKey`, the underlying `Signature` would only be initialized for signing, and not verification.

### DIFF
--- a/src/main/java/net/schmizz/sshj/signature/Signature.java
+++ b/src/main/java/net/schmizz/sshj/signature/Signature.java
@@ -22,17 +22,6 @@ import java.security.PublicKey;
 public interface Signature {
 
     /**
-     * Initialize this signature with the given public key and private key. If the private key is null, only signature
-     * verification can be performed.
-     *
-     * @param pubkey (null-ok) specify in case verification is needed
-     * @param prvkey (null-ok) specify in case signing is needed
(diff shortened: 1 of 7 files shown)


hierynomus:
Looks good.

charlesrgould:
Is there anything I can do to fix the test failure under Oracle JDK 9?

hierynomus:
I hope I just fixed it on master. Let's see what travis thinks :)

src/main/java/net/schmizz/sshj/signature/Signature.java
```diff
@@ -27,10 +27,32 @@
      *
      * @param pubkey (null-ok) specify in case verification is needed
      * @param prvkey (null-ok) specify in case signing is needed
+     * @deprecated Use {@link #initVerify(PublicKey)} or {@link #initSign(PrivateKey)} instead.
      */
+    @Deprecated
     void init(PublicKey pubkey, PrivateKey prvkey);
```
hierynomus: I don't think anybody has implemented their own signatures. And even if they did, we broke the contract already by adding the new methods. let's just remove the cruft and only have the new methods.

charlesrgould: Sure thing.

PR #368, opened 2017-10-04
Extracted ASN.1/DER encoding to method

charlesrgould:
Makes the ASN.1 encoding as similar as possible between SignatureDSA and SignatureECDSA.


hierynomus:
THanks for the cleanup (and additional test!)

PR #369, opened 2017-10-10
Migrate remaining block ciphers

charlesrgould:


src/main/java/com/hierynomus/sshj/transport/cipher/BlockCiphers.java
```diff
@@ -19,24 +19,46 @@
...
+        return new Factory(16, 192, "aes192-cbc", "AES", CIPHER_BLOCK_CHAINING_MODE);
+    }
+    public static Factory AES256CBC() {
+        return new Factory(16, 256, "aes256-cbc", "AES", CIPHER_BLOCK_CHAINING_MODE);
+    }
     public static Factory BlowfishCTR() {
         return new Factory(8, 256, "blowfish-ctr", "Blowfish", COUNTER_MODE);
     }
+    public static Factory BlowfishCBC() {
+        return new Factory(8, 128, "blowfish-cbc", "Blowfish", CIPHER_BLOCK_CHAINING_MODE);
```
charlesrgould: Should this be increased to 256 to match `BlowfishCTR`?

The value from `BlowfishCBC.java` was 16, which I multiplied by 8 (like the rest).

hierynomus: Yes, I think this should be the same as the `BlowfishCTR`

src/main/java/com/hierynomus/sshj/transport/cipher/BlockCiphers.java
```diff
@@ -19,24 +19,46 @@
...
+        return new Factory(16, 192, "aes192-cbc", "AES", CIPHER_BLOCK_CHAINING_MODE);
+    }
+    public static Factory AES256CBC() {
+        return new Factory(16, 256, "aes256-cbc", "AES", CIPHER_BLOCK_CHAINING_MODE);
+    }
     public static Factory BlowfishCTR() {
         return new Factory(8, 256, "blowfish-ctr", "Blowfish", COUNTER_MODE);
     }
+    public static Factory BlowfishCBC() {
+        return new Factory(8, 256, "blowfish-cbc", "Blowfish", CIPHER_BLOCK_CHAINING_MODE);
```
hierynomus: Oops, seems I was wrong:
(From: https://www.ietf.org/rfc/rfc4344.txt)

```
   The "blowfish-ctr" method uses Blowfish with 256-bit keys [SCHNEIER].
   The block size is 8 bytes.  (Note that "blowfish-cbc" from [RFC4253]
   uses 128-bit keys.)
```

So revert that commit... Sorry :)

charlesrgould: Nice find...

PR #372, opened 2017-10-10
Update net.i2p.crypto:eddsa to 0.2.0

paladox:


hierynomus:
Given that eddsa 0.2.0 is not dependent on Java8, I will merge this.

paladox:
thanks.

PR #406, opened 2018-03-09
Add support for multiple entries in known hosts

ragebiswas:
Issue#405 - similar to openssh, add support for attempting to verify
host keys by trying all entries in known_hosts, instead of bailing out
on the first failure. This is because there could be multiple entries for
a host, and only one/some of them may be valid.

Couple of notes:

- Used a generic `Exception` class in the `OpenSSHKnownHosts` constructor to continue with the rest of the file in case of other runtime exceptions (e.g. inability to base64 decode a line - covered by tests)
- Updated the `known_hosts` file to successfully validate multiple entries for the same host (where only one entry is valid)


hierynomus:
Did some refactoring on this PR, closing the original one.

ragebiswas:
Thanks @hierynomus  - a release would be great ;-)

PR #430, opened 2018-06-20
Fix possible OOME in ChannelInputStream

mstrap:
If the buffer is read slower than than incoming data arrives,
the buffer might continuing growing endlessly, finally resulting
in an OOME.


src/main/java/net/schmizz/sshj/connection/channel/ChannelInputStream.java
```diff
@@ -126,6 +126,24 @@ public void receive(byte[] data, int offset, int len)
...
+        // For slow readers, wait until the buffer has been completely read; this ensures that the buffer will be cleared
+        // in #read and the window position will be reset to 0. Otherwise, if the buffer is read slower than incoming data
+        // arrives, the buffer might continuing growing endlessly, finally resulting in an OOME.
+        // Note, that the buffer may still double its size once (provided that the maximum received chunk size is less
+        // than chan.getLocalMaxPacketSize).
+        for (; ; ) {
+            synchronized (buf) {
+                if (buf.wpos() >= chan.getLocalMaxPacketSize() && buf.available() > 0) {
+                    buf.notifyAll();
+                    Thread.yield();
```
hierynomus: Why would the `yield` be appropriate here? As the documentation specifies this is not something you typically call. Furthermore this is a busy wait construct, which is not pretty... Can't we solve this with a lock or latch.
Without a load of tests I'll also not gladly accept this PR..

mstrap: We could replace the yield by a very short sleep as a 32K buffer becomes quickly filled when having a high bandwidth -- maybe 1ms? To me, this seemed the most simple solution, but I'm quite new to SSHJ.

hierynomus: That is exactly the reason why I think that limiting it on the current starting buffer size is a bad idea. We should limit it on some configurable amount, else you get a very staggered pattern... 

Did you experience something in production or is this something hypothetical...?

mstrap: > That is exactly the reason why I think that limiting it on the current starting buffer size is a bad idea. We should limit it on some configurable amount, else you get a very staggered pattern...

That's true; I'm actually seeing this kind of pattern.

> Did you experience something in production or is this something hypothetical...?

We are getting such OOME's reported quite frequently from our users (Git client) and I'm perfectly able to reproduce this when cloning a Git repository from my local VM.

hierynomus:
@mstrap Can you work on the comments?

mstrap:
@hierynomus , so your suggestions are to:

1. Introduce something like `Channel.getLocalMaxQueueSize` and use instead of `getLocalMaxPacketSize`
2. Replace the `Thread.yield` by a short `Thread.sleep()` -- something like 1ms

? I'm happy to follow up with a patch, however without any additional tests.

mstrap:
This is most likely resolved by #913. We will test for a while without this patch and return if the problem recurs.

PR #458, opened 2018-10-22
Improving logging for KeyExchanger

Andremoniy:
In the `KeyExchanger` class there is the following error logging:

`
        log.error("Disconnecting because none of the configured Host key verifiers ({}) could verify '{}' host key with fingerprint {} for {}:{}",
                hostVerifiers,
                KeyType.fromKey(key),
                SecurityUtils.getFingerprint(key),
                transport.getRemoteHost(),
                transport.getRemotePort());
`

However the `hostVerifiers`'s elements do not properly override the `toString()` method.

It would be nice to see actual contents of this array for debug purposes.

### DIFF
--- a/src/main/java/net/schmizz/sshj/transport/KeyExchanger.java
+++ b/src/main/java/net/schmizz/sshj/transport/KeyExchanger.java
@@ -40,7 +40,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 final class KeyExchanger
         implements SSHPacketHandler, ErrorNotifiable {
 
-    private static enum Expected {
+    private enum Expected {
         /** we have sent or are sending KEXINIT, and expect the server's KEXINIT */
         KEXINIT,
         /** we are expecting some followup data as part of the exchange */
--- a/src/main/java/net/schmizz/sshj/transport/KeyExchanger.java
+++ b/src/main/java/net/schmizz/sshj/transport/KeyExchanger.java
@@ -40,7 +40,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 final class KeyExchanger
         implements SSHPacketHandler, ErrorNotifiable {
 
-    private enum Expected {
+    private static enum Expected {
         /** we have sent or are sending KEXINIT, and expect the server's KEXINIT */
         KEXINIT,
         /** we are expecting some followup data as part of the exchange */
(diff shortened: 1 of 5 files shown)


hierynomus:
@Andremoniy Do you have any idea how many entries a typical `known_hosts` file contains? I have 96 in there. If we would add this logging, you would not have a readable log file.

Andremoniy:
> @Andremoniy Do you have any idea how many entries a typical `known_hosts` file contains? I have 96 in there. If we would add this logging, you would not have a readable log file.

Fair enough. However the error itself says about some critical situation which has to be investigated. Say we have this issue on a bamboo server, so we loose actual docker container contents but want to check what's going on based on the logs. 

Would you approve a compromise solution with log.debug(...) printing out the contents of the hostVerifier's list?

Andremoniy:
> @Andremoniy Do you have any idea how many entries a typical `known_hosts` file contains? I have 96 in there. If we would add this logging, you would not have a readable log file.

P.S. I don't really think that a list of 96 default object identifiers is somehow better than their actual contents :p

hierynomus:
> > @Andremoniy Do you have any idea how many entries a typical `known_hosts` file contains? I have 96 in there. If we would add this logging, you would not have a readable log file.
> 
> P.S. I don't really think that a list of 96 default object identifiers is somehow better than their actual contents :p
> With great respect to your great library 👍

You'd now only get a single OID ;) (`OpenSshKnownHosts@129839...`), whereas with the current change you get one very big string (with 96 lines)....

I agree that a `toString` for each `HostKeyVerifier` improves the error message. But I would suggest to only print the  Classname and the filename that backs the `OpenSSHKnownHosts` verifier. So for instance `OpenSSHKnownHosts[khFile=/home/user/.ssh/known_hosts]`, so not the actual entries in the known_hosts files.

src/main/java/net/schmizz/sshj/transport/KeyExchanger.java
```diff
@@ -40,7 +40,7 @@
 final class KeyExchanger
         implements SSHPacketHandler, ErrorNotifiable {
 
-    private static enum Expected {
+    private enum Expected {
```
hierynomus: Not sure why this change is in there ;)

Andremoniy: Oups, I am sorry. I did it mechanically because IDEA highlighted it as redundant modifier (which is redundant indeed). Since it is certainly is unrelated changed, I have removed it :)

PR #472, opened 2018-11-26
Handle server initiated global requests

pepijnve:
RFC 4254 section 4 states that either side of the connection
may send global requests at any time and that the receiver
MUST respond appropriately.

This PR replaces the default logic of responding with 'UNIMPLEMENTED'
with the correct 'MSG_REQUEST_FAILURE' if 'wantReply' is set.


hierynomus:
Thanks @pepijnve!

pepijnve:
Context for this change is that I kept seeing `Sending SSH_MSG_UNIMPLEMENTED` debug messages scroll by which led me to wonder what wasn't implemented exactly.
Turns out on of the servers I was connected to was sending `keep-alive@openssh.com` global requests periodically. After this change those get handled silently.

hierynomus:
Thanks for the background info!

Issue #440, opened 2018-07-31
[PROTOCOL_ERROR] Received data for nonexistent channel 1.


tancolo:
Thank you for your amazing lib.
sshj version: 0.26.0 (not use source code)
Client: Android

I am facing a strange problem: In the process of using sshj, the SSH tunnel is successfully established, but in the process of passing data, an error occurs:
```
net.schmizz.sshj.transport.TransportException: [PROTOCOL_ERROR] Received data for nonexistent channel 1.
	at net.schmizz.sshj.transport.TransportImpl.gotDisconnect(TransportImpl.java:548)
	at net.schmizz.sshj.transport.TransportImpl.handle(TransportImpl.java:508)
	at net.schmizz.sshj.transport.Decoder.decode(Decoder.java:102)
	at net.schmizz.sshj.transport.Decoder.received(Decoder.java:170)
	at net.schmizz.sshj.transport.Reader.run(Reader.java:59)
```
This will cause the SSH connection to be broken.

So, my question is:
-  "[PROTOCOL_ERROR] Received data for nonexistent channel 1"  What is the underlying cause of this error? 

I noticed that someone reported a similar problem in 2016: issue(https://github.com/hierynomus/sshj/issues/274)

hierynomus:
Hi @tancolo,
Can you get me a full trace level log from initial connect up to the exception?

tancolo:
hello @hierynomus ,
From the debug log of my project,  I could not see the logs about sshj,  just see that error as above. 
If you want the full level log about your sshj lib,   how to open the log switcher on your sshj lib? 

BTW,  did you use the SLF4J in your sshj?  From my debug log, it shows some System.error, like this:
```
W/System.err( 8782): SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
W/System.err( 8782): SLF4J: Defaulting to no-operation (NOP) logger implementation
W/System.err( 8782): SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

Thank you.

hierynomus:
Yes, sshj uses slf4j as a logging api. Try to add logback as a logging impl to the classpath. Dont forget to add an logback xml file to enable trace logging.

tancolo:
hello @hierynomus,
I got the logs,  the log file is a bit large size. so I send it to you email. (jeroen@hierynomus.com) 
The title is [PROTOCOL_ERROR] Received data for nonexistent channel 1. #440
[...]

PR #476, opened 2018-12-05
Fix the issue 440

tancolo:
Hello @hierynomus 
I sent you a few emails about issue #440, an analysis and a solution provided, and how to use the docker cotainer to reproduce the issue.

So, I created a PR, please reivew it.
Thank you!


hierynomus:
This is now fixed. Thanks!

Issue #478, opened 2018-12-06
EdDSAPublicKeySpec NPE


solind:
Using the latest SSHJ/master, I ran into the following while connecting to an Ubuntu 18.04 box:

```
net.schmizz.sshj.transport.TransportException: null
        at net.schmizz.sshj.transport.TransportException$1.chain(TransportException.java:33)
        at net.schmizz.sshj.transport.TransportException$1.chain(TransportException.java:27)
        at net.schmizz.concurrent.Promise.deliverError(Promise.java:95)
        at net.schmizz.concurrent.Event.deliverError(Event.java:74)
        at net.schmizz.concurrent.ErrorDeliveryUtil.alertEvents(ErrorDeliveryUtil.java:34)
        at net.schmizz.sshj.transport.KeyExchanger.notifyError(KeyExchanger.java:392)
        at net.schmizz.sshj.transport.TransportImpl.die(TransportImpl.java:600)
        at net.schmizz.sshj.transport.Reader.run(Reader.java:65)
Caused by: net.schmizz.sshj.common.SSHException: null
        at net.schmizz.sshj.common.SSHException$1.chain(SSHException.java:36)
        at net.schmizz.sshj.common.SSHException$1.chain(SSHException.java:29)
        at net.schmizz.sshj.transport.TransportImpl.die(TransportImpl.java:595)
        at net.schmizz.sshj.transport.Reader.run(Reader.java:65)
Caused by: java.lang.NullPointerException: null
        at net.i2p.crypto.eddsa.spec.EdDSAPublicKeySpec.<init>(EdDSAPublicKeySpec.java:31)
        at net.schmizz.sshj.common.KeyType$6.readPubKeyFromBuffer(KeyType.java:187)
        at net.schmizz.sshj.common.Buffer.readPublicKey(Buffer.java:468)
        at net.schmizz.sshj.transport.kex.AbstractDHG.next(AbstractDHG.java:66)
        at net.schmizz.sshj.transport.KeyExchanger.handle(KeyExchanger.java:364)
        at net.schmizz.sshj.transport.TransportImpl.handle(TransportImpl.java:503)
        at net.schmizz.sshj.transport.Decoder.decodeMte(Decoder.java:159)
        at net.schmizz.sshj.transport.Decoder.decode(Decoder.java:79)
        at net.schmizz.sshj.transport.Decoder.received(Decoder.java:231)
        at net.schmizz.sshj.transport.Reader.run(Reader.java:59)
```

[...]

PR #489, opened 2019-01-30
Remove hard dependency on BouncyCastle

solind:
I deactivate certain KeyType enum members when BouncyCastle is not registered, to prevent ClassNotFoundExceptions.

Also, I updated a log.error call in KeyExchanger to permit use with SFL4J 1.6.2.

Solves issue #478.


solind:
Hey @hierynomus I'd love to add coverage for this change so that it would pass through the codecov hurdle, but ... I honestly have no idea how.  Ideally we'd need a way to run a test WITHOUT the BouncyCastle dependency in the classpath.  Care to provide me with a hint?

hierynomus:
There are only `available` methods, which aren't being called anywhere. This PR seems incomplete still.
Furthermore, there are more places where the logging uses the SLF4j formatting, instead of `String.format`. I do not want to change that, `String.format` incurs a penalty, as that eagerly renders the formatted string.

solind:
I eliminated String.format.  The ```available``` methods are called in the ```fromKey``` method, in KeyType line 409.

solind:
I'm going to close this.  When we're done integrating the last set of changes, we'll create a new PR.

PR #500, opened 2019-03-27
Making OpenSSHKnownHosts.EntryFactory public

dimitrovvlado:
The constructor of `OpenSSHKnownHosts.EntryFactory` should be public to enable creating in-memory known_hosts files. Currently if anyone decides to implement a `HostKeyVerifier` and reuse the parsing logic from `OpenSSHKnownHosts.KnownHostEntry`, he/she cannot do so because of the package-private constructor.

I've added an example on how this can be used for presentation purposes (I can exclude the example from the PR).


hierynomus:
Merged manually

PR #337, opened 2017-07-06
Add support for tunneling TCP/IP connections.

liff:
Can be used for connecting via jump servers, for example.

Should also solve #324 and #310 I believe.

It's rather simple but I'm not sure if the implementation in SocketClient is completely peachy. I'll gladly change it if there's a better approach.

Issue #324, opened 2017-04-27
SSH Proxy support


juddgaddie:
Is there any support or planned support for ssh proxying, for example with openssh:

```
ssh -o ProxyCommand="ssh -W %h:%p -q user@proxy" user@server

juddgaddie:
Closing as its 3 years old

PR #511, opened 2019-05-10
add support for tunneling via channels

aiwaniuk:
This is similar to https://github.com/hierynomus/sshj/pull/337
but this previous PR doesn't solve this issue: https://github.com/hierynomus/sshj/issues/324

This should allow doing this by passing or session.exec() session.startSubsystem() to this new method.


aiwaniuk:
@hierynomus could you merge this?

PR #513, opened 2019-05-13
fix exception when parsing max possible date

aiwaniuk:
I'm trying to add support for ssh-rsa-cert-v01@openssh.com for host certificates.
In our setup this cert valid before the date set to 18446744073709551615 (which is max possible date). In this case, cert parsing code throws an exception because readUInt64  cand handle values > Long.MAX_VALUE


hierynomus:
Can you add a test for this cornercase?

aiwaniuk:
test added

PR #514, opened 2019-05-14
Added support for RSA certificates as host key

aiwaniuk:
Added support to host verification using ssh-rsa-cert-v01@openssh.com host_key


hierynomus:
Can you add a test with a certificate?

aiwaniuk:
added test and removed dates check (I think this should be done at host_keyt verify phase)

aiwaniuk:
@hierynomus any chance for review/merging it and two other PRs?

hierynomus:
Yes, will try to have a look next week.

PR #520, opened 2019-07-16
Minor updates

TobiX:
This updates the Gradle Wrapper to latest Gradle 4 version (deprecations need to be handled before updating further), since I had some problems with the old Gradle version in my companies network.

Additionally, this updates most test dependencies to the latest version. This doesn't seem to break anything for me 😅 (This also gets rid of the extra Maven repository, since the newer Mockito is available from Maven Central)


TobiX:
I'm a bit surprised by the coverage change since I didn't touch the code at all. Maybe newer JaCoCo version?

hierynomus:
Thanks for the pr

PR #524, opened 2019-08-01
Forgive redundant spaces in OpenSSHv2 public keys and known_hosts

werehuman:
Sometimes users copy-pastes private and public keys in text editors. It leads to redundant spaces
and newlines. OpenSSH can easily read such keys, so users expect from SSHJ the same.


werehuman:
Sorry, it took a time for me to understand that I broke down working functionality. Bugs was fixed, build passed.

PR #535, opened 2019-09-12
Fix non-ASCII passwords

werehuman:
According to [RFC 4252](https://tools.ietf.org/html/rfc4252#section-8) and [RFC 4256](https://tools.ietf.org/html/rfc4256#section-3.4), passwords should be encoded in UTF-8. Currently all `char`s in passwords are being rounded to `byte` with loosing precision, so any attempt to authenticate with a non-ASCII symbol in a password fails.


hierynomus:
Thanks!

PR #547, opened 2019-11-20
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`


src/main/java/net/schmizz/sshj/sftp/Response.java
```diff
@@ -29,7 +29,7 @@
         FAILURE(4),
         BAD_MESSAGE(5),
         NO_CONNECTION(6),
-        CONNECITON_LOST(7),
+        CONNECTION_LOST(7),
```
jsoref: This is possibly an API break

hierynomus: It might indeed break the API, though the Response packet is typically not exposed. But let's remove it from the PR

jsoref:
@hierynomus 
I'm not sure why the test failed the second time w/ that commit dropped. It passed when I used my own Travis, seems like a flaky test.

https://travis-ci.org/jsoref/sshj/builds/614671580

PR #552, opened 2020-01-21
Use nano time instead of current time millis.

matenhagen:
Relative time should be measured by System.nanoTime instead of System.currentTimeMillis

currentTimeMillis might return incorrect results during time adjustments, for example day light savings adjustments.


hierynomus:
Thanks!

PaulaNadiaNatan (approved):

PaulaNadiaNatan (changes requested):
Karol łęczycki

Issue #553, opened 2020-01-22
Sending data to stdin of remote process - closing has no effect ->remote process waits indefinitely 


jpstotz:
Hi, i am trying to send some data to the stdin of the remote process via sshj. 
The sending itself is not a big problem (using `Command.getOutputStream() `), however some commands do not terminate until stdin is closed. In seems like it is impossible to use such commands via sshj as closing the OutputStream doe snot has any effect on the remote stdin.

I created the following simple example to demonstarte the problem.

It executes remotely the command `cat` which simply prints everything on stdout it receives on stdin. Via sshj I am sending the data `"Line1\nLine2\nLine3\n"` to it and I see that this works as the listening thread prints the received lines correctly. 
However in the end this always end in an `ConnectionException: Timeout expired` as the remote `cat` process does not terminate because stdin is not closed.

What do I have to send to close the stdin on remote side after the sent data has been received and processed by `cat` (note this just an example. I want to send larger data to a tar process to extract it on remote side without saving the tar file first)?

		try (Session session = ssh.startSession()) {
			final Command cmd = session.exec("cat");

			Thread t = new Thread() {

				@Override
				public void run() {
					try {
						InputStream in = cmd.getInputStream();
						BufferedReader br = new BufferedReader(new InputStreamReader(in));
						String line;
						while ((line = br.readLine()) != null) {
							System.out.println("## " + line);
						}
					} catch (Exception e) {
						e.printStackTrace();
					}
				}
			};
			t.start();

			byte[] data = "Line1\nLine2\nLine3\n".getBytes();
			OutputStream out = cmd.getOutputStream();
			out.write(data);
			out.flush();
			out.close(); // this should close stdin on remote side but it doesn't do

			// cmd.getInputStream()
			cmd.join(5, TimeUnit.SECONDS);
			System.out.println("\n** exit status: " + cmd.getExitStatus());
		} catch (Exception e) {
			e.printStackTrace();
		} finally {
			ssh.disconnect();
		}

jpstotz:
Based on the commented out line in ChannelOutputStream.close() I got some new ideas:

https://github.com/hierynomus/sshj/blob/master/src/main/java/net/schmizz/sshj/connection/channel/ChannelOutputStream.java#L166

Git blame has pointed me to #143 in which this line was commented out, reverted, commented out again and so on. In the end #143 has been closed with that line still commented out (what I don't understand as the issue is therefore IMHO not solved). 

[...]

Issue #496, opened 2019-03-18
scmd and EOF


grimdug:
Seems like newer sshj versions do not transmit an EOF to the running command. This did work fine at least up to version 0.8.1 and now just hangs (runs into timeout) with recent versions. I force the EOF by closing the cmd.getOutputStream() after all data was sent using a custom StreamCopier.:

//
// Do something like:  echo "hello" | ssh myhost cat
//
class MyStreamCopier {
  private final InputStream in;
  private final OutputStream out;
  MyStreamCopier(InputStream in, OutputStream out) {
    this.in = in;
    this.out = out;
  }
  public void spawn() {
    Thread thread = new Thread() {
      @Override
      public void run() {
        try {
          final byte[] buf = new byte[5];
          int read;
          while((read = in.read(buf)) != -1) {
            out.write(buf, 0, read);
            out.flush();
          }
          out.flush();
          out.close();
        } catch(IOException ioe) {
          System.err.println(ioe);
        }
      }
    };
    thread.start();
  }
}
class SSH {
  public static void main(String... args) throws IOException {
    final SSHClient ssh = new SSHClient();
    ssh.addHostKeyVerifier((String hostname, int port, PublicKey key) -> true);
    ssh.connect("myhost");
    try {
      ssh.authPassword(System.getProperty("user.name"), "mypass".toCharArray());
      try(Session session = ssh.startSession()) {
        Session.Command cmd = session.exec("cat");
        InputStream instream = new ByteArrayInputStream("hello\n".getBytes(StandardCharsets.UTF_8));
        new StreamCopier(cmd.getInputStream(), System.out).spawn("stdout");
        new StreamCopier(cmd.getErrorStream(), System.err).spawn("stderr");
        //new StreamCopier(cmd.getInputStream(), System.out, LoggerFactory.DEFAULT).spawn("stdout");
        //new StreamCopier(cmd.getErrorStream(), System.err, LoggerFactory.DEFAULT).spawn("stderr");
        new MyStreamCopier(instream, cmd.getOutputStream()).spawn();
        cmd.join(10, TimeUnit.SECONDS);
      }
    } finally {
      ssh.disconnect();
    }
  }
}

hierynomus:
I think this should now be closed with eb09a16

Issue #143, opened 2014-08-24
Unable to notify remote exec command of an EOF of the stdin


tsingakbar:
I'm trying to use sshj to implement something like `cat foo.tar.gz | ssh user@remote tar xzf -`, and I used the `Command.getOutputStream()` to feed these bytes from a local file named foo.tar.gz, but when all bytes is fed, I can't find a way to shutdown the process gracefully. 
Invoking the close() method on the stream won't work.

hierynomus:
@tsingakbar @nstogner Could one of you verify that that fixes the problem, and write a unit test that shows this solves it? Then I'll readd the line. I don't know why it was removed, the commit message isn't really saying much... @shikhar Why was it removed?

shikhar:
I was trying to simplify the channel close flow, and thought sending the EOF was redundant since in typical usage the channel would be closed soon after.

hierynomus:
Commit d955865 has been reverted in commit 703a0df. This should fix this issue.

hierynomus:
I'm commenting the line out again, just ran into the following:

```
2015-01-20 15:22:36.578 [TaskExecutionEngine-akka.actor.default-dispatcher-3]  DEBUG n.s.sshj.connection.ConnectionImpl - Attaching `session` channel (#0)
2015-01-20 15:22:36.731 [reader]  DEBUG n.s.s.c.c.direct.SessionChannel - Initialized - < session channel: id=0, recipient=0, localWin=[winSize=2097152], remoteWin=[winSize=0] >
2015-01-20 15:22:36.731 [TaskExecutionEngine-akka.actor.default-dispatcher-3]  DEBUG n.s.s.c.c.direct.SessionChannel - Will request to exec `ls -ld /tmp/ot-20150120T152236244`
2015-01-20 15:22:36.731 [TaskExecutionEngine-akka.actor.default-dispatcher-3]  DEBUG n.s.s.c.c.direct.SessionChannel - Sending channel request for `exec`
2015-01-20 15:22:36.732 [reader]  DEBUG n.s.s.c.c.direct.SessionChannel - Received window adjustment for 2097152 bytes
2015-01-20 15:22:36.733 [reader]  DEBUG n.s.s.c.channel.Window$Remote - Increasing by 2097152 up to 2097152
2015-01-20 15:22:36.736 [reader]  DEBUG n.s.s.c.c.direct.SessionChannel - Got chan request for `exit-status`
[...]

PR #554, opened 2020-01-22
Allow sending CHANNEL_EOF message when ChannelOutputStream is closed

jpstotz:
As discussed in #553 it is sometimes necessary to send an CHANNEL_EOF message while in other situation it is redundant. 

Therefore the best solution in my opinion is to let the programmer choose if to send the CHANNEL_EOF message or not. Until now sending such a message requires some ugly hacks using reflection which is really bad. Therefore I implemented a new property that allows to enable sending an CHANNEL_EOF message when `close()` is called.

It is designed to be used in this way:

	final Command cmd = session.exec("cat");
	ChannelOutputStream out = (ChannelOutputStream) cmd.getOutputStream();
	out.setSignalCloseToRemote(true);
	out.write(data);
	out.close();

Note: I did not provide any additional unit tests as it is pretty complicated to write unit test that uses this feature and works on all common platforms (Linux, Window,s OSX, ...) as it requires an command to be executed that waits for stdin like `cat` or `tar x`. However not all platforms provide these commands by default, therefore the test would fail on at least one platform.

### DIFF
--- a/src/main/java/net/schmizz/sshj/connection/channel/ChannelOutputStream.java
+++ b/src/main/java/net/schmizz/sshj/connection/channel/ChannelOutputStream.java
@@ -37,6 +37,7 @@ public final class ChannelOutputStream extends OutputStream implements ErrorNoti
     private final byte[] b = new byte[1];
 
     private boolean closed;
+    private boolean signalCloseToRemote = false;
     private SSHException error;
 
     private final class DataBuffer {
@@ -163,7 +164,9 @@ public final class ChannelOutputStream extends OutputStream implements ErrorNoti
         if (!closed) {
             try {
                 buffer.flush(false);
-//                trans.write(new SSHPacket(Message.CHANNEL_EOF).putUInt32(chan.getRecipient()));
+                if (signalCloseToRemote) {
+                    trans.write(new SSHPacket(Message.CHANNEL_EOF).putUInt32(chan.getRecipient()));
+                }
             } finally {
                 closed = true;
             }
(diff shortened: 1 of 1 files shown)


jpstotz:
@hierynomus from my perspective this PR is now ready - feel free to comment or to criticize it.

As I wrote in my first post I don't see a good way to increase the code coverage because of the problems providing a platform independent unit test for this scenario.

jpstotz:
If I understand it correctly also bug #496 is also about the inability to send the CHANNEL_EOF when the OutputStream is closed. Therefore by applying this PR this would make it possible "fix" this issue, too.

hierynomus:
@jpstotz I'm hesitant to merge this in, as I think the reason I commented it out is that there is a race condition with closing sometimes. I would rather always send the EOF on the channel, which is allowed according to the spec. See in #143 for when this got commented out.

jpstotz:
I am not sure what you mean with "race condition". The log of #143 does not contain many detailed error description. To me it is just a bunch of messages with works, works not, works again, works not... 

The only detailed description is [your post](https://github.com/hierynomus/sshj/issues/143#issuecomment-70799592) - but if I understand it correctly this error happened when the "line" was commented out - which means that no EOF was send, correct?

So the main question remains: Do you see the chance for a general solution? 

Because otherwise this workaround would at least to allow users to send EOF manually and therefore enable a lot of functionality which is currently unusable.

hierynomus:
I've run a small test connecting 1000 times and doing a command. I've decided to re-implement the fix in #143 unconditionally, and if there is a new bug, we'll have a look at it. So this PR can be closed I think.

PR #586, opened 2020-05-26
Make KeyType compatible with Android Keystore

FabianHenneke:
Android Keystore private keys do not implement PrivateKey since the
raw key material is not available to applications.

With this commit, sshj's KeyType correctly detects the algorithm
associated with Android Keystore keys, which makes them usable for SSH
authentication.

### DIFF
--- a/src/main/java/net/schmizz/sshj/common/ECDSAVariationsAdapter.java
+++ b/src/main/java/net/schmizz/sshj/common/ECDSAVariationsAdapter.java
@@ -103,7 +103,7 @@ class ECDSAVariationsAdapter {
     }
 
     static boolean isECKeyWithFieldSize(Key key, int fieldSize) {
-        return "ECDSA".equals(key.getAlgorithm())
+        return ("ECDSA".equals(key.getAlgorithm()) || "EC".equals(key.getAlgorithm()))
                 && fieldSizeFromKey((ECKey) key) == fieldSize;
     }
 
--- a/src/main/java/net/schmizz/sshj/common/ECDSAVariationsAdapter.java
+++ b/src/main/java/net/schmizz/sshj/common/ECDSAVariationsAdapter.java
@@ -15,6 +15,7 @@
  */
 package net.schmizz.sshj.common;
 
+import com.hierynomus.sshj.common.KeyAlgorithm;
 import com.hierynomus.sshj.secg.SecgUtils;
 import org.bouncycastle.asn1.nist.NISTNamedCurves;
 import org.bouncycastle.asn1.x9.X9ECParameters;
@@ -87,7 +88,7 @@ class ECDSAVariationsAdapter {
             ECPoint p = new ECPoint(bigX, bigY);
             ECPublicKeySpec publicKeySpec = new ECPublicKeySpec(p, ecCurveSpec);
 
-            KeyFactory keyFactory = KeyFactory.getInstance("ECDSA");
+            KeyFactory keyFactory = KeyFactory.getInstance(KeyAlgorithm.ECDSA);
             return keyFactory.generatePublic(publicKeySpec);
         } catch (Exception ex) {
             throw new GeneralSecurityException(ex);
@@ -103,7 +104,7 @@ class ECDSAVariationsAdapter {
     }
 
     static boolean isECKeyWithFieldSize(Key key, int fieldSize) {
-        return ("ECDSA".equals(key.getAlgorithm()) || "EC".equals(key.getAlgorithm()))
+        return (KeyAlgorithm.ECDSA.equals(key.getAlgorithm()) || KeyAlgorithm.EC_KEYSTORE.equals(key.getAlgorithm()))
                 && fieldSizeFromKey((ECKey) key) == fieldSize;
     }
 
--- a/src/main/java/net/schmizz/sshj/common/KeyType.java
+++ b/src/main/java/net/schmizz/sshj/common/KeyType.java
@@ -66,7 +66,7 @@ public enum KeyType {
 
         @Override
         protected boolean isMyType(Key key) {
-            return (key instanceof RSAPublicKey || key instanceof RSAPrivateKey);
+            return "RSA".equals(key.getAlgorithm());
         }
     },
 
@@ -99,7 +99,7 @@ public enum KeyType {
 
         @Override
         protected boolean isMyType(Key key) {
-            return (key instanceof DSAPublicKey || key instanceof DSAPrivateKey);
+            return "DSA".equals(key.getAlgorithm());
         }
 
     },
--- a/src/main/java/net/schmizz/sshj/common/KeyType.java
+++ b/src/main/java/net/schmizz/sshj/common/KeyType.java
@@ -15,6 +15,7 @@
  */
 package net.schmizz.sshj.common;
 
+import com.hierynomus.sshj.common.KeyAlgorithm;
 import com.hierynomus.sshj.signature.Ed25519PublicKey;
 import com.hierynomus.sshj.userauth.certificate.Certificate;
 import net.i2p.crypto.eddsa.EdDSAPublicKey;
@@ -30,9 +31,7 @@ import java.security.GeneralSecurityException;
 import java.security.Key;
 import java.security.KeyFactory;
 import java.security.PublicKey;
-import java.security.interfaces.DSAPrivateKey;
 import java.security.interfaces.DSAPublicKey;
-import java.security.interfaces.RSAPrivateKey;
 import java.security.interfaces.RSAPublicKey;
 import java.security.spec.DSAPublicKeySpec;
 import java.security.spec.RSAPublicKeySpec;
@@ -53,7 +52,7 @@ public enum KeyType {
             } catch (Buffer.BufferException be) {
                 throw new GeneralSecurityException(be);
             }
-            final KeyFactory keyFactory = SecurityUtils.getKeyFactory("RSA");
+            final KeyFactory keyFactory = SecurityUtils.getKeyFactory(KeyAlgorithm.RSA);
             return keyFactory.generatePublic(new RSAPublicKeySpec(n, e));
         }
 
@@ -66,7 +65,7 @@ public enum KeyType {
 
         @Override
         protected boolean isMyType(Key key) {
-            return "RSA".equals(key.getAlgorithm());
+            return KeyAlgorithm.RSA.equals(key.getAlgorithm());
         }
     },
 
@@ -84,7 +83,7 @@ public enum KeyType {
             } catch (Buffer.BufferException be) {
                 throw new GeneralSecurityException(be);
             }
-            final KeyFactory keyFactory = SecurityUtils.getKeyFactory("DSA");
+            final KeyFactory keyFactory = SecurityUtils.getKeyFactory(KeyAlgorithm.DSA);
             return keyFactory.generatePublic(new DSAPublicKeySpec(y, p, q, g));
         }
 
@@ -99,7 +98,7 @@ public enum KeyType {
 
         @Override
         protected boolean isMyType(Key key) {
-            return "DSA".equals(key.getAlgorithm());
+            return KeyAlgorithm.DSA.equals(key.getAlgorithm());
         }
 
     },
--- a/src/main/java/com/hierynomus/sshj/common/KeyAlgorithm.java
+++ b/src/main/java/com/hierynomus/sshj/common/KeyAlgorithm.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright (C)2009 - SSHJ Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.hierynomus.sshj.common;
 
 public class KeyAlgorithm {
--- a/src/main/java/com/hierynomus/sshj/common/KeyAlgorithm.java
+++ b/src/main/java/com/hierynomus/sshj/common/KeyAlgorithm.java
@@ -1,0 +1,16 @@
+package com.hierynomus.sshj.common;
+
+public class KeyAlgorithm {
+
+    public static final String RSA = "RSA";
+    public static final String DSA = "DSA";
+
+    /** Elliptic curve signature key algorithm for use with BouncyCastle **/
+    public static final String ECDSA = "ECDSA";
+
+    /** General elliptic curve algorithm identifier for use with BouncyCastle **/
+    public static final String EC_BC = "EC";
+
+    /** General elliptic curve algorithm identifier for use with the Android Keystore **/
+    public static final String EC_KEYSTORE = "EC";
+}
--- a/src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
+++ b/src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
@@ -15,6 +15,7 @@
  */
 package com.hierynomus.sshj.userauth.keyprovider;
 
+import com.hierynomus.sshj.common.KeyAlgorithm;
 import com.hierynomus.sshj.transport.cipher.BlockCiphers;
 import net.i2p.crypto.eddsa.EdDSAPrivateKey;
 import net.i2p.crypto.eddsa.spec.EdDSANamedCurveTable;
@@ -207,7 +208,7 @@ public class OpenSSHKeyV1KeyFile extends BaseFileKeyProvider {
                 keyBuffer.readMPInt(); // iqmp (q^-1 mod p)
                 keyBuffer.readMPInt(); // p (Prime 1)
                 keyBuffer.readMPInt(); // q (Prime 2)
-                kp = new KeyPair(publicKey, SecurityUtils.getKeyFactory("RSA").generatePrivate(new RSAPrivateKeySpec(n, d)));
+                kp = new KeyPair(publicKey, SecurityUtils.getKeyFactory(KeyAlgorithm.RSA).generatePrivate(new RSAPrivateKeySpec(n, d)));
                 break;
             case ECDSA256:
                 kp = new KeyPair(publicKey, createECDSAPrivateKey(kt, keyBuffer, "P-256"));
@@ -239,7 +240,7 @@ public class OpenSSHKeyV1KeyFile extends BaseFileKeyProvider {
         X9ECParameters ecParams = NISTNamedCurves.getByName(name);
         ECNamedCurveSpec ecCurveSpec = new ECNamedCurveSpec(name, ecParams.getCurve(), ecParams.getG(), ecParams.getN());
         ECPrivateKeySpec pks = new ECPrivateKeySpec(s, ecCurveSpec);
-        return SecurityUtils.getKeyFactory("ECDSA").generatePrivate(pks);
+        return SecurityUtils.getKeyFactory(KeyAlgorithm.ECDSA).generatePrivate(pks);
 
     }
 }
--- a/src/main/java/net/schmizz/sshj/transport/kex/Curve25519DH.java
+++ b/src/main/java/net/schmizz/sshj/transport/kex/Curve25519DH.java
@@ -15,6 +15,7 @@
  */
 package net.schmizz.sshj.transport.kex;
 
+import com.hierynomus.sshj.common.KeyAlgorithm;
 import net.schmizz.sshj.common.Factory;
 import net.schmizz.sshj.transport.random.Random;
 import org.bouncycastle.asn1.x9.X9ECParameters;
@@ -31,7 +32,7 @@ public class Curve25519DH extends DHBase {
     private byte[] secretKey;
 
     public Curve25519DH() {
-        super("ECDSA", "ECDH");
+        super(KeyAlgorithm.ECDSA, "ECDH");
     }
 
     @Override
--- a/src/main/java/net/schmizz/sshj/transport/kex/ECDH.java
+++ b/src/main/java/net/schmizz/sshj/transport/kex/ECDH.java
@@ -15,6 +15,7 @@
  */
 package net.schmizz.sshj.transport.kex;
 
+import com.hierynomus.sshj.common.KeyAlgorithm;
 import net.schmizz.sshj.common.Factory;
 import net.schmizz.sshj.common.SecurityUtils;
 import net.schmizz.sshj.transport.random.Random;
@@ -54,7 +55,7 @@ public class ECDH extends DHBase {
 
     @Override
     public void computeK(byte[] f) throws GeneralSecurityException {
-        KeyFactory keyFactory = SecurityUtils.getKeyFactory("EC");
+        KeyFactory keyFactory = SecurityUtils.getKeyFactory(KeyAlgorithm.EC_BC);
         ECPublicKeySpec keySpec = new ECPublicKeySpec(getDecoded(f, ecParameterSpec.getCurve()), ecParameterSpec);
         PublicKey yourPubKey = keyFactory.generatePublic(keySpec);
         agreement.doPhase(yourPubKey, true);
--- a/src/main/java/net/schmizz/sshj/transport/verification/OpenSSHKnownHosts.java
+++ b/src/main/java/net/schmizz/sshj/transport/verification/OpenSSHKnownHosts.java
@@ -15,6 +15,7 @@
  */
 package net.schmizz.sshj.transport.verification;
 
+import com.hierynomus.sshj.common.KeyAlgorithm;
 import com.hierynomus.sshj.transport.verification.KnownHostMatchers;
 import net.schmizz.sshj.common.*;
 import org.slf4j.Logger;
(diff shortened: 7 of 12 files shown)


src/main/java/net/schmizz/sshj/common/KeyType.java
```diff
@@ -66,7 +66,7 @@ protected void writePubKeyContentsIntoBuffer(PublicKey pk, Buffer<?> buf) {
 
         @Override
         protected boolean isMyType(Key key) {
-            return (key instanceof RSAPublicKey || key instanceof RSAPrivateKey);
+            return "RSA".equals(key.getAlgorithm());
```
hierynomus, line 69: Let's extract `"RSA"` and `"DSA"` into constants.

src/main/java/net/schmizz/sshj/common/ECDSAVariationsAdapter.java
```diff
@@ -103,7 +103,7 @@ static void writePubKeyContentsIntoBuffer(PublicKey pk, Buffer<?> buf) {
     }
 
     static boolean isECKeyWithFieldSize(Key key, int fieldSize) {
-        return "ECDSA".equals(key.getAlgorithm())
+        return ("ECDSA".equals(key.getAlgorithm()) || "EC".equals(key.getAlgorithm()))
```
hierynomus, line 106: Not sure where `"EC"` is coming from? Might want to add a test to cover that branch if that's possible.

hierynomus: While you're changing it, also extract `"ECDSA"` into a constant, not sure why I didn't do that.

hierynomus:
I've merged master into this branch to fix the builds.

src/main/java/net/schmizz/sshj/common/KeyType.java
FabianHenneke: I created a new class `KeyAlgorithm` in `common` that extracts all the constants used (implicitly) as arguments to `KeyFactory.getInstance(...)`.

src/main/java/net/schmizz/sshj/common/ECDSAVariationsAdapter.java
FabianHenneke: The extraction should be done with the latest commit and it also contains a comment that explains where the "EC" comes from - it is the standard algorithm identifier for everything EC-related in the Android Keystore. Unfortunately, testing this seems impossible when not on the Android platform.

FabianHenneke:
I closed and re-opened to see whether this would trigger travis. Unfortunately, it didn't.

hierynomus:
> I closed and re-opened to see whether this would trigger travis. Unfortunately, it didn't.

I still need to remove travis :) Now with GH Actions travis is no longer needed...

FabianHenneke:
Thanks! Are you able to provide an estimate fow when 0.30.0 will be released?

oakkitten:
i can't get this to work.

my device is running Android 10. in gradle i'm using

    api 'com.hierynomus:sshj:0.30.0'
    implementation 'org.bouncycastle:bcpkix-jdk15on:1.65'

i tried "importing" BC in a variety of ways, and stuff below are best case scenarios

* first, i can't import any keys into Android Key Store (AKS).

    this `KeyPair` created by [sshlib](https://github.com/connectbot/sshlib/blob/main/src/main/java/com/trilead/ssh2/crypto/PEMDecoder.java) is importable into AKS:

    ```java
    KeyPair kp = PEMDecoder.decode(charKey, sshPassword);
    ```

    this `KeyPair` created by sshj, while usable in `authPublickey()`, is not:

    ```java
    KeyFormat keyFormat = KeyProviderUtil.detectKeyFileFormat(
            new InputStreamReader(new ByteArrayInputStream(sshKey)), false);
    final FileKeyProvider fileKeyProvider = Factory.Named.Util.create(
            new DefaultConfig().getFileKeyProviderFactories(), keyFormat.toString());
    if (fileKeyProvider == null)
        throw new SSHException("No provider available for " + keyFormat + " key file");
    fileKeyProvider.init(new InputStreamReader(new ByteArrayInputStream(sshKey)),
            PasswordUtils.createOneOff(sshPassword.toCharArray()));
    KeyPair kp = new KeyPair(fileKeyProvider.getPublic(), fileKeyProvider.getPrivate());
    ```

    public keys are the same in both cases, but sshlib creates `RSA Private CRT Key` instead of `RSA Private Key`. when [creating a certificate](https://github.com/ubergeek42/weechat-android/blob/8fdb31946ed66755fdd393a606552cbd45ebd822/app/src/main/java/com/ubergeek42/WeechatAndroid/utils/AndroidKeyStoreUtils.java#L49-L97) to import this into AKS, i'm getting:

    ```
    PrivateKeyPickerPreference: Error while putting "RSA" key into AndroidKeyStore
        java.lang.ArithmeticException: BigInteger division by zero
            at java.math.NativeBN.BN_div(Native Method)
            at java.math.BigInt.division(BigInt.java:312)
            at java.math.BigInteger.remainder(BigInteger.java:955)
            at org.bouncycastle.crypto.engines.RSACoreEngine.processBlock(Unknown Source:28)
            at org.bouncycastle.crypto.engines.RSABlindedEngine.processBlock(Unknown Source:54)
            at org.bouncycastle.crypto.encodings.PKCS1Encoding.encodeBlock(Unknown Source:76)
            at org.bouncycastle.crypto.encodings.PKCS1Encoding.processBlock(Unknown Source:4)
            at org.bouncycastle.crypto.signers.RSADigestSigner.generateSignature(Unknown Source:25)
            at org.bouncycastle.operator.bc.BcSignerOutputStream.getSignature(Unknown Source:2)
            at org.bouncycastle.operator.bc.BcContentSignerBuilder$1.getSignature(Unknown Source:2)
            at org.bouncycastle.cert.X509v3CertificateBuilder.generateSig(Unknown Source:12)
            at org.bouncycastle.cert.X509v3CertificateBuilder.build(Unknown Source:40)
            at com.ubergeek42.WeechatAndroid.utils.AndroidKeyStoreUtils.generateCertificate(AndroidKeyStoreUtils.java:94)
            ...
    ```

    certificates for EC keys can be generated (i added `case "ECDSA:`), but can't be [imported into AKS](https://github.com/ubergeek42/weechat-android/blob/8fdb31946ed66755fdd393a606552cbd45ebd822/app/src/main/java/com/ubergeek42/WeechatAndroid/utils/AndroidKeyStoreUtils.java#L148-L155). I assume that it's because the key algorithm is "ECDSA" instead of "EC":

    ```
    java.security.KeyStoreException: Unsupported key algorithm: ECDSA
        at android.security.keystore.AndroidKeyStoreSpi.getLegacyKeyProtectionParameter(AndroidKeyStoreSpi.java:348)
        at android.security.keystore.AndroidKeyStoreSpi.setPrivateKeyEntry(AndroidKeyStoreSpi.java:360)
        at android.security.keystore.AndroidKeyStoreSpi.engineSetKeyEntry(AndroidKeyStoreSpi.java:294)
        at java.security.KeyStore.setKeyEntry(KeyStore.java:1179)
        at com.ubergeek42.WeechatAndroid.utils.AndroidKeyStoreUtils.putKeyPairIntoAndroidKeyStore(AndroidKeyStoreUtils.java:153)
    ```

    i didn't try DSA.

* second, i cannot connect (using keys from AKS)

    using RSA key, i get the AndroidKeyStoreRSAPrivateKey is not a RSAPrivateKey (which it isn't):

    ```
    net.schmizz.sshj.userauth.UserAuthException: Exhausted available authentication methods
        at net.schmizz.sshj.SSHClient.auth(SSHClient.java:227)
        at net.schmizz.sshj.SSHClient.authPublickey(SSHClient.java:342)
        at net.schmizz.sshj.SSHClient.authPublickey(SSHClient.java:360)
        at com.ubergeek42.weechat.relay.connection.SSHConnection.connect(SSHConnection.java:134)
        at com.ubergeek42.weechat.relay.connection.RelayConnection.lambda$connect$1$RelayConnection(RelayConnection.java:90)
        at com.ubergeek42.weechat.relay.connection.-$$Lambda$RelayConnection$CgdZlWpK5gNUVzAJypQrciXI8lg.run(Unknown Source:2)
        at com.ubergeek42.weechat.relay.connection.RelayConnection$Protected.run(RelayConnection.java:180)
        at com.ubergeek42.weechat.relay.connection.Events$EventStream.lambda$new$1$Events$EventStream(Events.java:60)
        at com.ubergeek42.weechat.relay.connection.-$$Lambda$Events$EventStream$GPjMZUunGsQtYUQNi0r5b_jZ9qc.run(Unknown Source:2)
        at com.ubergeek42.weechat.relay.connection.Utils$FriendlyThread.run(Utils.java:108)
     Caused by: net.schmizz.sshj.userauth.UserAuthException: Supplied key (android.security.keystore.AndroidKeyStoreRSAPrivateKey) is not a RSAPrivateKey instance
        at net.schmizz.sshj.userauth.UserAuthException$1.chain(UserAuthException.java:33)
        at net.schmizz.sshj.userauth.UserAuthException$1.chain(UserAuthException.java:26)
        at net.schmizz.concurrent.Promise.deliverError(Promise.java:95)
        at net.schmizz.sshj.userauth.UserAuthImpl.notifyError(UserAuthImpl.java:157)
        at net.schmizz.sshj.transport.TransportImpl.die(TransportImpl.java:620)
        at net.schmizz.sshj.transport.Reader.run(Reader.java:66)
     Caused by: net.schmizz.sshj.common.SSHException: Supplied key (android.security.keystore.AndroidKeyStoreRSAPrivateKey) is not a RSAPrivateKey instance
        at net.schmizz.sshj.common.SSHException$1.chain(SSHException.java:36)
        at net.schmizz.sshj.common.SSHException$1.chain(SSHException.java:29)
        at net.schmizz.sshj.transport.TransportImpl.die(TransportImpl.java:614)
        at net.schmizz.sshj.transport.Reader.run(Reader.java:66) 
     Caused by: net.schmizz.sshj.common.SSHRuntimeException: Supplied key (android.security.keystore.AndroidKeyStoreRSAPrivateKey) is not a RSAPrivateKey instance
        at net.schmizz.sshj.signature.AbstractSignature.initSign(AbstractSignature.java:68)
        at net.schmizz.sshj.userauth.method.KeyedAuthMethod.putSig(KeyedAuthMethod.java:79)
        at net.schmizz.sshj.userauth.method.AuthPublickey.sendSignedReq(AuthPublickey.java:74)
        at net.schmizz.sshj.userauth.method.AuthPublickey.handle(AuthPublickey.java:45)
        at net.schmizz.sshj.userauth.UserAuthImpl.handle(UserAuthImpl.java:143)
        at net.schmizz.sshj.transport.TransportImpl.handle(TransportImpl.java:515)
        at net.schmizz.sshj.transport.Decoder.decodeMte(Decoder.java:159)
        at net.schmizz.sshj.transport.Decoder.decode(Decoder.java:79)
        at net.schmizz.sshj.transport.Decoder.received(Decoder.java:231)
        at net.schmizz.sshj.transport.Reader.run(Reader.java:60)
     Caused by: java.security.InvalidKeyException: Supplied key (android.security.keystore.AndroidKeyStoreRSAPrivateKey) is not a RSAPrivateKey instance
        at org.bouncycastle.jcajce.provider.asymmetric.rsa.DigestSignatureSpi.engineInitSign(Unknown Source:50)
        at java.security.Signature$Delegate.engineInitSign(Signature.java:1383)
        at java.security.Signature.initSign(Signature.java:679)
        at net.schmizz.sshj.signature.AbstractSignature.initSign(AbstractSignature.java:66)
        at net.schmizz.sshj.userauth.method.KeyedAuthMethod.putSig(KeyedAuthMethod.java:79) 
        at net.schmizz.sshj.userauth.method.AuthPublickey.sendSignedReq(AuthPublickey.java:74) 
        at net.schmizz.sshj.userauth.method.AuthPublickey.handle(AuthPublickey.java:45) 
        at net.schmizz.sshj.userauth.UserAuthImpl.handle(UserAuthImpl.java:143) 
        at net.schmizz.sshj.transport.TransportImpl.handle(TransportImpl.java:515) 
        at net.schmizz.sshj.transport.Decoder.decodeMte(Decoder.java:159) 
        at net.schmizz.sshj.transport.Decoder.decode(Decoder.java:79) 
        at net.schmizz.sshj.transport.Decoder.received(Decoder.java:231) 
        at net.schmizz.sshj.transport.Reader.run(Reader.java:60) 
    ```

    using EC key, i get

    ```
    ...
    Caused by: java.security.InvalidKeyException: cannot identify EC private key: java.security.InvalidKeyException: no encoding for EC private key
        at org.bouncycastle.jcajce.provider.asymmetric.util.ECUtil.generatePrivateKeyParameter(Unknown Source:225)
        at org.bouncycastle.jcajce.provider.asymmetric.ec.SignatureSpi.engineInitSign(Unknown Source:0)
        at java.security.Signature$Delegate.engineInitSign(Signature.java:1383)
        at java.security.Signature.initSign(Signature.java:679)
        at net.schmizz.sshj.signature.AbstractSignature.initSign(AbstractSignature.java:66)
    ...
    ```

    this also [checks if it's ECPrivateKey](https://github.com/bcgit/bc-java/blob/master/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/util/ECUtil.java#L229) and fails as [the encoded key is null](https://github.com/bcgit/bc-java/blob/master/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/util/ECUtil.java#L269)

any ideas as to what might be wrong?

perhaps we can have a tiny example of Android + sshj + AKS?

FabianHenneke:
> * first, i can't import any keys into Android Key Store (AKS).

I would strongly recommend to generate keys within the Android Keystore, not import keys from the outside. If you want to protect an existing SSH key, consider wrapping it using the AndroidX Security library's `EncryptedFile`. For an example of how this is done with an ed25519 key in the context of SSHJ, see https://github.com/android-password-store/Android-Password-Store/blob/8a4316067396e6a0a5490e727197cf843d3f4ec6/app/src/main/java/com/zeapo/pwdstore/git/sshj/SshKey.kt#L217-L253.

Because I have never tried this myself, I unfortunately can't help you with these issues.

> * second, i cannot connect (using keys from AKS)

This I might be able to help you with. The stack traces you provided seem to indicate that SSHJ is specifying the engine for its signature operations to be specifically BC, whereas the Keystore-backed keys require a different engine that should not be specified explicitly. With sshj 0.30.0, you should be able to work around this in the following way:

1. Loop over all security providers looking for Android's own "BC" implementation and remove it.
2. Insert the Java BC provider at that position.
3. Instruct SSHJ to not register BC itself and specify a `null` `SecurityProvider`.

I implemented these steps here: https://github.com/android-password-store/Android-Password-Store/blob/8a4316067396e6a0a5490e727197cf843d3f4ec6/app/src/main/java/com/zeapo/pwdstore/git/sshj/SshjConfig.kt#L36-L57

oakkitten:
thanks for trying to help!

> strongly recommend to generate keys within the Android Keystore

it's a good idea, and we might implement it some day (low priority), but the ability to import existing keys is something our users expect so it better work

> wrapping it using the AndroidX Security library

also a good idea, thanks. as i understand, the key material must be decrypted into memory in order for the key to perform actual crypto? but yeah, we should do it

> can't help you with the issues you are facing when importing keys

i think i solved this somewhat, [this code](https://github.com/ubergeek42/weechat-android/blob/ae9f30c1febec9533971e603024e67022b6130ac/app/src/main/java/com/ubergeek42/WeechatAndroid/utils/AndroidKeyStoreUtils.kt) loads pkcs1 and pksc8 files only relying on BC. the trick is using `JcaPEMKeyConverter()` instead of `JcaPEMKeyConverter().setProvider(bouncyCastleProvider)`
exclusively for pkcs1 operations (and maybe only for ECDSA operations), that gives us `EC` keys that are importable into AKS 

> This I might be able to help you with

i tried your code and i still can't connect, except if i use a very nasty trick

i tried running `setUpBouncyCastleForSshj()` verbatim (except different logging statement) from Application.onCreate and i'm getting

    JCE providers: AndroidNSSP (1.0), AndroidOpenSSL (1.0), CertPathProvider (1.0), AndroidKeyStoreBCWorkaround (1.0), BC (1.66), HarmonyJSSE (1.0), AndroidKeyStore (1.0)

and this error when connecting:

    net.schmizz.sshj.transport.TransportException: Unable to reach a settlement: [diffie-hellman-group1-sha1, diffie-hellman-group-exchange-sha1] and [curve25519-sha256, curve25519-sha256@libssh.org, ecdh-sha2-nistp256, ecdh-sha2-nistp384, ecdh-sha2-nistp521, diffie-hellman-group-exchange-sha256, diffie-hellman-group16-sha512, diffie-hellman-group18-sha512, diffie-hellman-group14-sha256, diffie-hellman-group14-sha1]
        at net.schmizz.sshj.transport.Proposal.firstMatch(Proposal.java:149)
        at net.schmizz.sshj.transport.Proposal.negotiate(Proposal.java:130)
        at net.schmizz.sshj.transport.KeyExchanger.gotKexInit(KeyExchanger.java:223)
        at net.schmizz.sshj.transport.KeyExchanger.handle(KeyExchanger.java:359)
        at net.schmizz.sshj.transport.TransportImpl.handle(TransportImpl.java:517)
        at net.schmizz.sshj.transport.Decoder.decodeMte(Decoder.java:159)
        at net.schmizz.sshj.transport.Decoder.decode(Decoder.java:79)
        at net.schmizz.sshj.transport.Decoder.received(Decoder.java:231)
        at net.schmizz.sshj.transport.Reader.run(Reader.java:60)

(on the sshd part, i get this:)

    Unable to negotiate with 192.168.1.128 port 41145: no matching key exchange method found. Their offer: diffie-hellman-group1-sha1,diffie-hellman-group-exchange-sha1 [preauth]

if i run my `sshd` as follows: `/usr/sbin/sshd -ddd -p 23 -oKexAlgorithms=+diffie-hellman-group1-sha1` then i can connect fine, but allowing something like `diffie-hellman-group1-sha1` sounds like a nasty nasty idea...

FabianHenneke:
> > wrapping it using the AndroidX Security library
> 
> also a good idea, thanks. as i understand, the key material must be decrypted into memory in order for the key to perform actual crypto? but yeah, we should do it

Yes, the contents of the decrypted file would be used to initialize a `PrivateKey`, which could then be passed into SSHJ. This was the only option for us as we wanted to support ed25519 keys which are not supported by the Android Keystore directly.

> 
> > can't help you with the issues you are facing when importing keys
> 
> i think i solved this somewhat, [this code](https://github.com/ubergeek42/weechat-android/blob/ae9f30c1febec9533971e603024e67022b6130ac/app/src/main/java/com/ubergeek42/WeechatAndroid/utils/AndroidKeyStoreUtils.kt) loads pkcs1 and pksc8 files only relying on BC. the trick is using `JcaPEMKeyConverter()` instead of `JcaPEMKeyConverter().setProvider(bouncyCastleProvider)`

This looks pretty good. The rule of thumb regarding the Android Keystore is that everything should work as long as no explicit provider is ever set anywhere.

> and this error when connecting:
> 
> ```
> net.schmizz.sshj.transport.TransportException: Unable to reach a settlement: [diffie-hellman-group1-sha1, diffie-hellman-group-exchange-sha1] and [curve25519-sha256, curve25519-sha256@libssh.org, ecdh-sha2-nistp256, ecdh-sha2-nistp384, ecdh-sha2-nistp521, diffie-hellman-group-exchange-sha256, diffie-hellman-group16-sha512, diffie-hellman-group18-sha512, diffie-hellman-group14-sha256, diffie-hellman-group14-sha1]
>     at net.schmizz.sshj.transport.Proposal.firstMatch(Proposal.java:149)
>     at net.schmizz.sshj.transport.Proposal.negotiate(Proposal.java:130)
>     at net.schmizz.sshj.transport.KeyExchanger.gotKexInit(KeyExchanger.java:223)
>     at net.schmizz.sshj.transport.KeyExchanger.handle(KeyExchanger.java:359)
>     at net.schmizz.sshj.transport.TransportImpl.handle(TransportImpl.java:517)
>     at net.schmizz.sshj.transport.Decoder.decodeMte(Decoder.java:159)
>     at net.schmizz.sshj.transport.Decoder.decode(Decoder.java:79)
>     at net.schmizz.sshj.transport.Decoder.received(Decoder.java:231)
>     at net.schmizz.sshj.transport.Reader.run(Reader.java:60)
> ```

You get this error because the SSHJ default config is not really usable without automatically registered BC. You will (and likely should) roll your own config anyway. Ours can be found [here](https://github.com/android-password-store/Android-Password-Store/blob/develop/app/src/main/java/com/zeapo/pwdstore/git/sshj/SshjConfig.kt#L197) for reference, just ignore the `OpenKeychainWrappedKeyAlgorithmFactory`.

oakkitten:
aha, i think i figured this part out! for the sake of testing, this works:

```java
SecurityUtils.setSecurityProvider("BC");
ssh = new SSHClient();
SecurityUtils.setSecurityProvider(null);
```

this will require a lot of work, still.

perhaps this would be best to put into `AndroidConfig`? that file seems outdated?

another small thing. how do i specify the types of host keys that i have for the server? looking at `Proposal.java` there's 

```java
firstMatch(this.getHostKeyAlgorithms(), other.getHostKeyAlgorithms()),
```

and host keys come from

```java
public List<String> getHostKeyAlgorithms() {
    return sig;
}
```

and `sig` comes from

```java
sig = Factory.Named.Util.getNames(config.getKeyAlgorithms());
```

but as key algorithms come from config, they are the same for every connection, aren't they? so if the server had only key of type RSA, and we remember it, and the server adds Ed25519 key, which has a higher priority in sshj, and we try to connect, we'll be getting that huge someone is doing something evil warning, won't we? instead of verifying using RSA

while it's possible to specify key algorithms in the config, it feels like a wrong place to do it?

also, i don't see how `OpenSSHKnownHosts` is doing this...

oakkitten:
also, ECDSA host keys will not work, as in `ECDSAVariationsAdapter.java` there's this piece of code:

```java
    static PublicKey readPubKeyFromBuffer(Buffer<?> buf, String variation) throws GeneralSecurityException {
        String algorithm = BASE_ALGORITHM_NAME + variation;
        if (!SecurityUtils.isBouncyCastleRegistered()) {
            throw new GeneralSecurityException("BouncyCastle is required to read a key of type " + algorithm);
        }
    ...
```

FabianHenneke:
> aha, i think i figured this part out! for the sake of testing, this works:
> 
> ```java
> SecurityUtils.setSecurityProvider("BC");
> ssh = new SSHClient();
> SecurityUtils.setSecurityProvider(null);
> ```
> 
> this will require a lot of work, still.
> 
> perhaps this would be best to put into `AndroidConfig`? that file seems outdated?
> 
> another small thing. how do i specify the types of host keys that i have for the server? looking at `Proposal.java` there's
> 
> ```java
> firstMatch(this.getHostKeyAlgorithms(), other.getHostKeyAlgorithms()),
> ```
> 
> and host keys come from
> 
> ```java
> public List<String> getHostKeyAlgorithms() {
>     return sig;
> }
> ```
> 
> and `sig` comes from
> 
> ```java
> sig = Factory.Named.Util.getNames(config.getKeyAlgorithms());
> ```
> 
> but as key algorithms come from config, they are the same for every connection, aren't they? so if the server had only key of type RSA, and we remember it, and the server adds Ed25519 key, which has a higher priority in sshj, and we try to connect, we'll be getting that huge someone is doing something evil warning, won't we? instead of verifying using RSA
> 

The negotiation algorithm is correct according to the relevant SSH RFC, but perhaps the client proposal, which you can tweak via the per-connection SshjConfig, should only consist of the subset of algorithms it already has a host key for (if not empty)? I don't even know how OpenSSH does this though. @hierynomus Do you have a recommendation on how to fix a host key type per connection?

> also, ECDSA host keys will not work, as in ECDSAVariationsAdapter.java there's this piece of code:

Good catch, that seems to be an Android compatibility issue. I will look into whether I can fix it.

oakkitten:
OpenSSH sends all keys, but it [reorders](https://github.com/openssh/openssh-portable/blob/master/sshconnect2.c#L122-L209) them [unless](https://github.com/openssh/openssh-portable/blob/master/sshconnect2.c#L250-L255) you [overwrite](https://github.com/openssh/openssh-portable/blob/master/sshconnect2.c#L226-L229) that in a config (see the penultimate paragraph):

    HostKeyAlgorithms
            Specifies the host key algorithms that the client wants to use in or‐
            der of preference.  Alternately if the specified value begins with a
            ‘+’ character, then the specified key types will be appended to the
            default set instead of replacing them.  If the specified value begins
            with a ‘-’ character, then the specified key types (including wild‐
            cards) will be removed from the default set instead of replacing them.
            The default for this option is:

               ecdsa-sha2-nistp256-cert-v01@openssh.com,
               ecdsa-sha2-nistp384-cert-v01@openssh.com,
               ecdsa-sha2-nistp521-cert-v01@openssh.com,
               ssh-ed25519-cert-v01@openssh.com,
               rsa-sha2-512-cert-v01@openssh.com,rsa-sha2-256-cert-v01@openssh.com,
               ssh-rsa-cert-v01@openssh.com,
               ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,
               ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa

            If hostkeys are known for the destination host then this default is
            modified to prefer their algorithms.

            The list of available key types may also be obtained using "ssh -Q
            key".

e.g. for a single RSA key,

    debug3: order_hostkeyalgs: prefer hostkeyalgs: rsa-sha2-512-cert-v01@openssh.com,rsa-sha2-256-cert-v01@openssh.com,ssh-rsa-cert-v01@openssh.com,rsa-sha2-512,rsa-sha2-256,ssh-rsa
    ...
    debug2: local client KEXINIT proposal
    ...
    debug2: host key algorithms: rsa-sha2-512-cert-v01@openssh.com,rsa-sha2-256-cert-v01@openssh.com,ssh-rsa-cert-v01@openssh.com,rsa-sha2-512,rsa-sha2-256,ssh-rsa,ecdsa-sha2-nistp256-cert-v01@openssh.com,ecdsa-sha2-nistp384-cert-v01@openssh.com,ecdsa-sha2-nistp521-cert-v01@openssh.com,ssh-ed25519-cert-v01@openssh.com,ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519

i'm assuming it sends all algos in order to in fact receive *some* host key so that it can print its hash to the user along with the warning

FabianHenneke:
I didn't know this is what OpenSSH is doing, but it should be easy to recreate in SSHJ: You can permute the host key algorithms in your instance of `SshjConfig` according to the memorized host key.

oakkitten:
this is doable, but feels hackish. requires me to recreate config which is supposed to be global, and to have an opinion on which host key algorithms are better. i'd expect something like `ssh.setAvailableHostKeyTypes(hostKeys.getAvailableHostKeyTypes(host, port))`

i'm not quite sure that you can get available host keys from `OpenSSHKnownHosts` without subclassing it

PR #587, opened 2020-05-27
Add pwdf retry logic to OpenSSHKeyV1KeyFile

FabianHenneke:
While PKCS8KeyFile uses PasswordFinder's shouldRetry to determine
whether it should call reqPassword again if decryption of they key file
fails, OpenSSHKeyV1KeyFile simply gives up and throws an exception.

With this commit, retry logic similar to that of PKCS8KeyFile is added
to OpenSSHKeyV1KeyFile. The PasswordFinder's reqPassword is called
again if the validation of the "checkint" fails, which indicates an
incorrect passphrase.

### DIFF
--- a/src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
+++ b/src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
@@ -28,7 +28,6 @@ import net.schmizz.sshj.userauth.keyprovider.KeyFormat;
 import org.bouncycastle.asn1.nist.NISTNamedCurves;
 import org.bouncycastle.asn1.x9.X9ECParameters;
 import org.bouncycastle.jce.spec.ECNamedCurveSpec;
-import org.bouncycastle.openssl.EncryptionException;
 import org.mindrot.jbcrypt.BCrypt;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -67,6 +66,13 @@ public class OpenSSHKeyV1KeyFile extends BaseFileKeyProvider {
         @Override
         public String getName() {
             return KeyFormat.OpenSSHv1.name();
+        }
+    }
+
+    private static class CheckIntsDifferedException extends IOException {
+
+        public CheckIntsDifferedException() {
+            super("The checkInts differed, the key was not correctly decoded. Passphrase may be incorrect.");
         }
     }
 
@@ -117,7 +123,7 @@ public class OpenSSHKeyV1KeyFile extends BaseFileKeyProvider {
                 PlainBuffer decrypted = decryptBuffer(decryptionBuffer, cipherName, kdfName, kdfOptions);
                 try {
                     return readUnencrypted(decrypted, publicKey);
-                } catch (EncryptionException e) {
+                } catch (CheckIntsDifferedException e) {
                     if (pwdf == null || !pwdf.shouldRetry(resource))
                         throw e;
                 }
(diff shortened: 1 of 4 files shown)


FabianHenneke:
I think that the CI fails here and in #586 are due to https://github.com/hierynomus/sshj/commit/35266945585b440ca75040d0af07181598531139. I confirmed that this builds and retries correctly locally, but some unit tests fail due to seemingly unrelated Mockito exceptions.

src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
```diff
@@ -28,6 +28,7 @@
 import org.bouncycastle.asn1.nist.NISTNamedCurves;
 import org.bouncycastle.asn1.x9.X9ECParameters;
 import org.bouncycastle.jce.spec.ECNamedCurveSpec;
+import org.bouncycastle.openssl.EncryptionException;
```
hierynomus, line 31: I'd rather not use an exception from BC.

hierynomus:
> I think that the CI fails here and in #586 are due to [3526694](https://github.com/hierynomus/sshj/commit/35266945585b440ca75040d0af07181598531139). I confirmed that this builds and retries correctly locally, but some unit tests fail due to seemingly unrelated Mockito exceptions.

The CI failures are indeed due to that change. Have to dive into it why this happens.

src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
FabianHenneke: I tried to stay as close as possible to https://github.com/hierynomus/sshj/blob/b1be9258b4f22257f5c45cabc9566e6702beb8b3/src/main/java/net/schmizz/sshj/userauth/keyprovider/PKCS8KeyFile.java#L84, but am open to any other type. Should I just define a new Exception for this purpose as an inner class of OpenSSHKeyV1KeyFile?

FabianHenneke:
The second commit replaces `EncryptionException` with a new exception type defined as an inner class.

src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
hierynomus: Let's also change this occurrence with the same exception, to keep them in sync, which is definitely a good idea ;)

src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
```diff
@@ -111,8 +118,16 @@ private KeyPair readDecodedKeyPair(final PlainBuffer keyBuffer) throws IOExcepti
             return readUnencrypted(privateKeyBuffer, publicKey);
         } else {
             logger.info("Keypair is encrypted with: " + cipherName + ", " + kdfName + ", " + Arrays.toString(kdfOptions));
-            PlainBuffer decrypted = decryptBuffer(privateKeyBuffer, cipherName, kdfName, kdfOptions);
-            return readUnencrypted(decrypted, publicKey);
+            while (true) {
```
hierynomus, line 121: Can you also add a test for this?

hierynomus:
I've merged master into this PR, in order to fix the builds.

src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
FabianHenneke: I created a new `KeyDecryptionFailedException` and throw it from both `PKCS8KeyFile` and `OpenSSHKeyV1KeyFile`.

src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
FabianHenneke: I added tests that either fail or pass after a number of retries.

hierynomus:
Thanks!

PR #589, opened 2020-06-02
Introduce KeyAlgorithm to separate KeyType from Algorithm (Fixes #588)

### LINKED ISSUE #588: Support for SHA-256/512 with ssh-rsa keys
It would be great if support for the modern [`rsa-sha2-256` and `rsa-sha2-512` signature types](https://tools.ietf.org/html/rfc8332) could be added. Currently, RSA keys are limited to SHA1.

I would work on this myself, but I do not know how to implement this cleanly. The current design seems to be based on the assumption that there is a 1-to-1 mapping between key types and signature types. If you can provide me with a rough sketch of how this could be fit into the code, I would execute the plan when I have time.

hierynomus:


hierynomus:
This PR also enables ssh-rsa2-256 and ssh-rsa2-512 signatures

Issue #590, opened 2020-06-05
Make available server extension data


dkocher:
It should be possible to query server extensions such as `statvfs`.

hierynomus:
PR is merged!

PR #591, opened 2020-06-05
Allow to query for server extensions available.

dkocher:
Fix for #590.

### DIFF
--- a/src/main/java/net/schmizz/sshj/sftp/SFTPEngine.java
+++ b/src/main/java/net/schmizz/sshj/sftp/SFTPEngine.java
@@ -108,6 +108,14 @@ public class SFTPEngine
         return operativeVersion;
     }
 
+    public boolean supportsServerExtension(final String extension, final String domain) {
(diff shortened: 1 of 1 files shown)


hierynomus:
Shouldn't we also be checking the server extensions when doing extended requests?

dkocher:
> Shouldn't we also be checking the server extensions when doing extended requests?

I think that should be the responsibility of the user. We use it like 

```
if supportsServerExtension("statvfs", "openssh.com")
    newExtendedRequest("statvfs@openssh.com")
````

hierynomus:
> > Shouldn't we also be checking the server extensions when doing extended requests?
> 
> I think that should be the responsibility of the user. We use it like
> 
> ```
> if supportsServerExtension("statvfs", "openssh.com")
>     newExtendedRequest("statvfs@openssh.com")
> ```

It would actually be good to guard against sending bad requests to the server, as we know which extended requests are available. I'll add that in a separate PR and merge this one, as it is a pre-existing situation.

Issue #592, opened 2020-06-05
Allow to query for allowed authentication methods


dkocher:
Currently `Iterable` is defined in the interface which does not allow to query for methods.

hierynomus:
PR is merged ;)

PR #593, opened 2020-06-05
Review interface to return collection for allowed authentication meth…

dkocher:
…ods that allows to query for elements.

Fix for #592.


hierynomus:
Makes sense, I'll merge it once the pipeline's greeen

PR #596, opened 2020-06-05
Don't specify JcaPEM providers as null

FabianHenneke:
If no provider is set in the `SecurityUtils`, no named provider should be set for the `JcaPEMKeyConverter` as this would cause a `missing provider` exception. This currently breaks `PKCS8KeyFile` if `SecurityUtils.setSecurityProvider(null)` and `SecurityUtils.setRegisterBouncyCastle(false)` is used.


FabianHenneke:
I missed that there were actually two calls to `setProvider`. As far as I can tell, these are the only calls to `setProvider` in the codebase.

Issue #268, opened 2016-08-30
List files from remote directory fails with Serv-U


comtel2000:
Hi, if I try to list files from a remote directory (sample: export/files) I get a exception:

``` java
net.schmizz.sshj.sftp.SFTPException: No such file.
    at net.schmizz.sshj.sftp.Response.error(Response.java:113)
    at net.schmizz.sshj.sftp.Response.ensurePacketTypeIs(Response.java:91)
    at net.schmizz.sshj.sftp.SFTPEngine.openDir(SFTPEngine.java:153)
    at net.schmizz.sshj.sftp.SFTPClient.ls(SFTPClient.java:61)
    at net.schmizz.sshj.sftp.SFTPClient.ls(SFTPClient.java:56)
```

But only with Serv-U (tested with: Serv-U_15.1.4.6,Serv-U_15.1.3.43) and `SSHClient.newSFTPClient`
and with `StatefulSFTPClient sftp = new StatefulSFTPClient(engine.init());` it works very well.
I tested it also with other SFTP servers like CoreFTP, Windows embedded SFTP and both ways are ok.
Only with CoreFTP (mini-server 2.04) the SFTPEngine.close(); cause a TimeoutException all the time.

``` java
    try (SSHClient ssh = new SSHClient()) {
      ssh.addHostKeyVerifier(new PromiscuousVerifier());
      ssh.connect(host, port);
      ssh.authPassword(user, pwd);

      SFTPEngine engine = new SFTPEngine(ssh);
      StatefulSFTPClient sftp = new StatefulSFTPClient(engine.init());
      // auto close run into timeout exception with CoreFTP mini SFTP
      // try (StatefulSFTPClient sftp = new StatefulSFTPClient(engine.init())) {
      logger.info("found {} remote files", sftp.ls(incomingRemotePath).size());
    } catch (Exception e) {
      logger.error(e.getMessage(), e);
    }

    try (SSHClient ssh = new SSHClient()) {
      ssh.addHostKeyVerifier(new PromiscuousVerifier());
      ssh.connect(host, port);
      ssh.authPassword(user, pwd);

      SFTPClient sftp = ssh.newSFTPClient();
      // newSFTPClient doesn't work with Serv-U SFTP
      logger.info("found {} remote files", sftp.ls(incomingRemotePath).size());
    } catch (Exception e) {
      logger.error(e.getMessage(), e);
    }

```

hierynomus:
Which version of SSHJ are you using?

[...]

PR #603, opened 2020-06-22
Added method to created Stateful clients ( #268).

dajoropo:
New method to create Stateful clients similar to the standard ones to avoid the issues with Serv-U servers as described here: https://github.com/hierynomus/sshj/issues/268#issuecomment-647403159


hierynomus:
Approved, thanks for the addition.

PR #607, opened 2020-06-23
Fix matching of pubkeys to key algorithms

### LINKED ISSUE #600: Public key authentication only possible with host key types
Since 9671352bdaaf13f00681866c09332b1e88a113fd, it seems that public key authentication can only succeed if the key type is also a host key type supported by the server.

I believe that the root cause of this is https://github.com/hierynomus/sshj/commit/9671352bdaaf13f00681866c09332b1e88a113fd#diff-521b24d9ed9094555c19f59ffa14e862R235-R239, where the "signature algorithms" (i.e. `server_host_key_algorithms` per the SSH spec) are used to populate the factory that is [later used to send and sign with public keys](https://github.com/hierynomus/sshj/commit/9671352bdaaf13f00681866c09332b1e88a113fd#diff-c112c7a229db1b0a9b377c3c9e0bafa5R71-R77).

I would think that that the list of `KeyAlgorithm` factories should rather be populated directly from the config as it does not require negotiation with the server. `KeyedAuthMethod.java` would then need to be changed to use those factories instead of asking `Transport` for it.

FabianHenneke:
Allow all configured key algorithms for pubkey authentication, even if
these algorithms are not supported as host key algorithms by the
server.

Preference is given to the modern rsa-sha2-* signature algorithms if
the server indicates support for them as host keys signature
algorithms.

Should fix #600.


hierynomus:
Is this according to the RFC? This code was changed pretty recently and _I think_ I then interpreted the RFC correctly. But I might be mistaken of course.

FabianHenneke:
> Is this according to the RFC? This code was changed pretty recently and _I think_ I then interpreted the RFC correctly. But I might be mistaken of course.

I am quite sure that pubkey algorithms are not negotiated, only host key algorithms are. There is an extension with which the server can announce support for pubkey types though which could be more reliable than the rough feature detection implemented here based on the host key type.

src/main/java/net/schmizz/sshj/transport/Proposal.java
```diff
@@ -127,13 +128,14 @@ public NegotiatedAlgorithms negotiate(Proposal other)
             throws TransportException {
         return new NegotiatedAlgorithms(
                 firstMatch(this.getKeyExchangeAlgorithms(), other.getKeyExchangeAlgorithms()),
-                allMatch(this.getSignatureAlgorithms(), other.getSignatureAlgorithms()),
+                firstMatch(this.getHostKeyAlgorithms(), other.getHostKeyAlgorithms()),
```
hierynomus, line 131: I think we need to do a bit more magic here.... See RFC4253:

> kex_algorithms
         Key exchange algorithms were defined above.  The first
         algorithm MUST be the preferred (and guessed) algorithm.  If
         both sides make the same guess, that algorithm MUST be used.
         Otherwise, the following algorithm MUST be used to choose a key
         exchange method: Iterate over client's kex algorithms, one at a
         time.  Choose the first algorithm that satisfies the following
         conditions:
           +  the server also supports the algorithm,
           +  if the algorithm requires an encryption-capable host key,
            there is an encryption-capable algorithm on the server's
            server_host_key_algorithms that is also supported by the
            client, and
           +  if the algorithm requires a signature-capable host key,
            there is a signature-capable algorithm on the server's
            server_host_key_algorithms that is also supported by the
            client.
      If no algorithm satisfying all these conditions can be found, the
      connection fails, and both sides MUST disconnect.

> server_host_key_algorithms
         A name-list of the algorithms supported for the server host
         key.  The server lists the algorithms for which it has host
         keys; the client lists the algorithms that it is willing to
         accept.  There MAY be multiple host keys for a host, possibly
         with different algorithms.
         Some host keys may not support both signatures and encryption
         (this can be determined from the algorithm), and thus not all
         host keys are valid for all key exchange methods.
         Algorithm selection depends on whether the chosen key exchange
         algorithm requires a signature or an encryption-capable host
         key.  It MUST be possible to determine this from the public key
         algorithm name.  The first algorithm on the client's name-list
         that satisfies the requirements and is also supported by the
         server MUST be chosen.  If there is no such algorithm, both
         sides MUST disconnect.

FabianHenneke: > I think we need to do a bit more magic here.... See RFC4253:
> 
> > kex_algorithms
> > Key exchange algorithms were defined above.  The first
> > algorithm MUST be the preferred (and guessed) algorithm.  If
> > both sides make the same guess, that algorithm MUST be used.
> > Otherwise, the following algorithm MUST be used to choose a key
> > exchange method: Iterate over client's kex algorithms, one at a
> > time.  Choose the first algorithm that satisfies the following
> > conditions:
> > +  the server also supports the algorithm,
> > +  if the algorithm requires an encryption-capable host key,
> > there is an encryption-capable algorithm on the server's
> > server_host_key_algorithms that is also supported by the
> > client, and
> > +  if the algorithm requires a signature-capable host key,
> > there is a signature-capable algorithm on the server's
> > server_host_key_algorithms that is also supported by the
> > client.
> > If no algorithm satisfying all these conditions can be found, the
> > connection fails, and both sides MUST disconnect.
> 
> > server_host_key_algorithms
> > A name-list of the algorithms supported for the server host
> > key.  The server lists the algorithms for which it has host
> > keys; the client lists the algorithms that it is willing to
> > accept.  There MAY be multiple host keys for a host, possibly
> > with different algorithms.
> > Some host keys may not support both signatures and encryption
> > (this can be determined from the algorithm), and thus not all
> > host keys are valid for all key exchange methods.
> > Algorithm selection depends on whether the chosen key exchange
> > algorithm requires a signature or an encryption-capable host
> > key.  It MUST be possible to determine this from the public key
> > algorithm name.  The first algorithm on the client's name-list
> > that satisfies the requirements and is also supported by the
> > server MUST be chosen.  If there is no such algorithm, both
> > sides MUST disconnect.

I agree that's true, but does SSHJ even implement a kex_algorithm that is encryption-based? Based on a quick glance they all seem to be signature-based, which would mean that the current behavior would be spec-compliant in practice. What do you think?

moritz31:
I face exactly this issue. Authentication happens over an ed25519 key but for signature the server only provides rsa, which will not work with the current implementation

src/main/java/net/schmizz/sshj/transport/Proposal.java
hierynomus: Yes, but _in practice_ does not mean it will be spec compliant forever. Let's try to adhere to the spec and validate. We might get an encryption based kex sooner or later...

Do you know which ones are encryption based?

FabianHenneke: I don't know of any encryption-based KEX algorithm, but will ask some people to make sure I'm not missing anything here.

FabianHenneke: I [asked for examples on StackExchange](https://security.stackexchange.com/questions/235276/examples-of-ssh-key-exchange-algorithms-requiring-encryption-capable-host-keys) but have not received any yet.

I also started reading the OpenSSH Portable source code and it seems that they also don't implement support for key exchanges that are not signature-based: In [kex_choose_conf](https://github.com/openssh/openssh-portable/blob/fec89f32a84fd0aa1afc81deec80a460cbaf451a/kex.c#L882), both [kex algorithm](https://github.com/openssh/openssh-portable/blob/fec89f32a84fd0aa1afc81deec80a460cbaf451a/kex.c#L819) and [host key algorithm](https://github.com/openssh/openssh-portable/blob/fec89f32a84fd0aa1afc81deec80a460cbaf451a/kex.c#L839) are independently determined by the equivalent of `firstMatch`.

Given that OpenSSH itself doesn't implement this part of the SSH spec and there don't seem to be any examples (not even the new post-quantum kex algorithm is one), I would say that this implementation of the abstract negotiation algorithm is fine as it is. If someone wanted to use an encryption-based kex algorithm with SSHJ, they likely would need to make other changes as well, including to the kex interface itself.

hierynomus: 👍 Will add a few small additional comments in the PR

FabianHenneke: Thanks!

FabianHenneke:
@moritz31 If possible, could you check whether the issue you are facing is resolved by my PR?

src/main/java/net/schmizz/sshj/transport/NegotiatedAlgorithms.java
```diff
@@ -15,37 +15,38 @@
...
     private final String c2sCipher;
     private final String s2cCipher;
     private final String c2sMAC;
     private final String s2cMAC;
     private final String c2sComp;
     private final String s2cComp;
 
-    NegotiatedAlgorithms(String kex, List<String> availableSigs, String c2sCipher, String s2cCipher, String c2sMAC, String s2cMAC,
-                         String c2sComp, String s2cComp) {
+    private final Boolean rsaSHA2Support;
```
hierynomus, line 29: Why is this not a primitive `boolean`?

src/main/java/net/schmizz/sshj/transport/TransportImpl.java
```diff
@@ -658,17 +661,31 @@ ConnInfo getConnInfo() {
...
 
-    public void setKeyAlgorithms(List<KeyAlgorithm> keyAlgorithms) {
-        this.keyAlgorithms = keyAlgorithms;
+    @Override
+    public void setRSASHA2Support(Boolean rsaSHA2Support) {
+        this.rsaSHA2Support = rsaSHA2Support;
+    }
+
+    @Override
+    public KeyAlgorithm getClientKeyAlgorithm(KeyType keyType) throws TransportException {
```
hierynomus, line 679: Can we add a test that failed with the old implementation and now succeeds. That will prevent regression

src/main/java/net/schmizz/sshj/transport/NegotiatedAlgorithms.java
FabianHenneke: No reason, I'm just not writing much Java these days. Fixed.

src/main/java/net/schmizz/sshj/transport/TransportImpl.java
FabianHenneke: I have a test, but it unearthed an unrelated issue with `ecdsa-sha2-nistp521` keys. I will see whether I can fix it.

FabianHenneke: I have added tests that fail with the old implementation and succeed with the new one. However, this requires https://github.com/hierynomus/sshj/pull/623, which fixes the incorrect encoding of long ECDSA signatures. Please merge that one first for the tests to pass.

src/main/java/net/schmizz/sshj/transport/Transport.java
```diff
@@ -238,5 +238,8 @@ long write(SSHPacket payload)
      */
     void die(Exception e);
 
-    KeyAlgorithm getKeyAlgorithm(KeyType keyType) throws TransportException;
+    void setHostKeyAlgorithm(KeyAlgorithm keyAlgorithm);
```
hierynomus, line 241: `KeyExchanger` has a `TransportImpl` to work with, so there is no need to expose the `set...` methods on the interface. Let's remove them..

FabianHenneke:
I'll make a minor improvement to the tests.

FabianHenneke:
> I'll make a minor improvement to the tests.

Nvm, everything is fine. It should just be noted that the SSH server in the docker image should not add  both `ecdsa-sha2-384` and `ecdsa-sha2-521` host keys, as otherwise the tests can't verify any longer that #600 has been fixed.

src/main/java/net/schmizz/sshj/transport/Transport.java
FabianHenneke: Done

FabianHenneke:
The flaky part seems to be the build step, not the integration test step. But everything passed now.

hierynomus:
yes, the plugins repository was throwing 500 errors...

PR #609, opened 2020-06-25
Clear passphrase bytes after use

FabianHenneke:
Mimics the behavior of `decrypt()` in `PKCS5KeyFile.java`.


hierynomus:
The behaviour is correct, but let's add a test that this doesn't get lost!

FabianHenneke:
I integrated checks for wiped passphrase bytes into the custom mock password finder I implemented for a previous PR. Note that the correct wiping of the intermediate buffers cannot be tested that way.

PR #623, opened 2020-07-28
Fix transport encoding of nistp521 signatures

FabianHenneke:
SignatureECDSA.encode() does not correctly handle signatures longer
than 128 bytes, which affects signatures using the nistp521 curve.

This commits fixes the issue by replacing the ad-hoc ASN.1 DER
parsing with a use of ASN1InputStream.


hierynomus:
Would be great to have a test here also :).

FabianHenneke:
> Would be great to have a test here also :).

The test will be added with #607. Unfortunately, due to #600, we could only add a test here if we were to add an `ecdsa-sha2-nistp521` host key, which then would mean that the tests in #607 would no longer verify that #600 has been fixed. This is a confusing situation, but once #623 and #607 have been merged (in this order), both will be covered by passing tests.

PR #628, opened 2020-09-03
fix keep-alive thread never be closed

myemperor:
TransportImpl have a default serivce as nullService, when you set keepAlive to retain connection, a keep-alive thread will be create and periodic send msg to remote. but even after sshClient.close() has been called. this keepAlive will still exist. 
So, the solution is to set a ConnectionImpl object to TransportImpl.


hierynomus:
I'm not sure I understand your comment, or why this should fix it. Can you elaborate? The fix doesn't feel correct...

dkocher:
Superseded by #752

PR #636, opened 2020-10-06
Improve Android compatibility

FabianHenneke:
* Instead of only counting BouncyCastle as being registered if it is set as the explicit security provider used by SSHJ, count it as
registered if it is available as a provider. This improves Android compatibility, which requires not specifying an explicit provider. This is meant to address https://github.com/hierynomus/sshj/pull/586#issuecomment-704006627 

@oakkitten Could you verify that this fixes the issue with ECDSA host keys?

* The ECNamedCurveGenParameterSpec is a BC-specific workaround for missing curve tables in Java 1.4 and earlier. For the sake of Android compatibility, where Conscrypt can't deal with this custom spec class, replace it with the standard ECGenParameterSpec and update the curve names to the standard identifiers.

### DIFF
--- a/src/main/java/net/schmizz/sshj/common/SecurityUtils.java
+++ b/src/main/java/net/schmizz/sshj/common/SecurityUtils.java
@@ -249,7 +249,14 @@ public class SecurityUtils {
      */
     public static synchronized boolean isBouncyCastleRegistered() {
         register();
-        return BOUNCY_CASTLE.equals(securityProvider) || SPONGY_CASTLE.equals(securityProvider);
+        Provider[] providers = Security.getProviders();
+        for (Provider provider : providers) {
+            String name = provider.getName();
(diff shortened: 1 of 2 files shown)


oakkitten:
i'll test the code later, but why check for BC in the first place? why not let it fail? same with the order places where the check for BC happens

oakkitten:
hm jitpack [doesn't seem to be able to build this](https://jitpack.io/com/github/FabianHenneke/sshj/397b77faddb01e47229361e637c6dad7dd15d720/build.log)

FabianHenneke:
Let's wait for @hierynomus to give his opinion. My first attempt would also have been to just remove all the explicit BC checks, but I do not know whether this has the potential to obscure bugs.

FabianHenneke:
I added another commit to fix failing ECDSA key exchanges on Android.

hierynomus (approved):

FabianHenneke:
Would it be possible to get a release with this fix and AES-GCM support?

PR #647, opened 2020-11-24
reduced log level for lenient handling of missing CR

hpoettker:
Currently, sshj logs a warning when a a server identifies itself with a string that is not terminated as expected with CR LF but only LF.

As a user of ssh, I may have no control over the server I am connecting to and can do nothing to remediate the underlying issue. At the same time, sshj handles this situation very gracefully. Therefore, I'd consider the log level _warning_ as too high.

To ensure helpful logging of the library, I suggest to reduce the log level from _warn_ to _info_.


hierynomus (approved):

hierynomus:
Thanks, agreed with the change.

Issue #563, opened 2020-02-24
SFTP Rename with flags (atomic)


ahweis:
The default SSH rename command do not sent any flags for how to handle the rename. In the SSH spec it is possible to send the three flags:

SSH_FXF_RENAME_OVERWRITE  0x00000001
SSH_FXF_RENAME_ATOMIC     0x00000002
SSH_FXF_RENAME_NATIVE     0x00000004

At the last uint32 part of the rename command. Will it be possible to extend the rename function to support a list of flags to include (default to 0) or create an overloaded function that supports this.

lucamilanesio:
@ahweis what happened to this issue? I provided a possible solution with the associated PR, can you have a look if that makes sense?

PR #652, opened 2020-12-08
Enable renaming with flags

lucamilanesio:
The SFTP protocol allows to rename files by specifying
extra flags:

- OVERWRITE
- ATOMIC
- NATIVE

The flags are exposed through a new RenameFlags enum and
can be passed as parameters to the rename() method in
SFTPClient/SFTPEngine.

Relates to #563


src/main/java/net/schmizz/sshj/sftp/RenameFlags.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (C)2009 - SSHJ Contributors
```
lucamilanesio, line 2: This should be 2020

src/main/java/net/schmizz/sshj/sftp/RenameFlags.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.schmizz.sshj.sftp;
+
+public enum RenameFlags {
+    OVERWRITE(0),
+    ATOMIC(1),
+    NATIVE(2);
```
lucamilanesio, line 22: These values are incorrect: they should be as follows.
```
      SSH_FXF_RENAME_OVERWRITE  0x00000001
       SSH_FXF_RENAME_ATOMIC     0x00000002
       SSH_FXF_RENAME_NATIVE     0x00000004
```

lucamilanesio:
@aperkins what do you think?

PR #680, opened 2021-04-13
Minor improvement for SCP preserve flag:

Estraysian:
**Problem:**
When trying to SCP a file to a remote machine, where the file exists (overwrite) and the user you're using for SCP is not the owner of the file, you get error "set mode: Operation not permitted", whenever -p switch is specified.
**Solution:**
Added a boolean variable to toggle usage of -p switch when using SCPUploadClient.copy()
Since it's an extend, the default behavior is unchanged, omitting -p is optional and needs to be explicitly specified.


hierynomus (approved):

hierynomus:
Thanks! Approved and merged, good addition.

Issue #686, opened 2021-05-12
Port forwarding: remote server closes connection, but sshj does not close connection to client


tsoiland:
TLDR; I have a failing unit test that I think points at a legitimate bug: [https://github.com/tsoiland/sshj/commit/4a7afbcc3ca14fc091a90f9faf74ce146c307328](https://github.com/tsoiland/sshj/commit/4a7afbcc3ca14fc091a90f9faf74ce146c307328)

I originally had a problem running docker exec (http api) through sshj. From what I've been able to gather it streams data back in the http response and probably doesn't know the content length when it writes the response headers, so the client has to wait until the connection is closed by the server. 

When I try to tunnel this through sshj the connection between my http client and sshj is never closed and therefore my http client waits forever.

This works with a direct connection to docker's http server.

It also works tunnelled through openssh-client and I can see the FIN, ACKs on both sides of the tunnel with wireshark.

vladimirlagunov:
Hi! It can be related to #143.

tsoiland:
1. I realized while doing some more debugging that my problem is with local port forwarding, not remote like in my linked pull request. New unit test with fix in StreamCopier.java: https://github.com/tsoiland/sshj/commit/54b2310b12d26c1638be50bb69f8b533dfea35c3

2. I looked at issue #143, but there the problem seems to be the other way around. Also uncommenting the suggested line doesn't help: `trans.write(new SSHPacket(Message.CHANNEL_EOF).putUInt32(chan.getRecipient()));`

PR #687, opened 2021-05-12
Failing unit test for issue #686

tsoiland:
Port forwarding: remote server closes connection, but sshj does not close connection to client


tsoiland:
To be clear: the latest commit also contains a fix, not just failing unit test.

hierynomus:
I'm not sure that this is actually a problem. You're trying to read a next response without actually doing a next request. The idea of the tunnels is that they're kept open for multiple requests afaik. So it should be possible to do a second GET request over the same tunnel. You can try that with OpenSSH also.

tsoiland:
I agree the tunnel should remain open. But the connection going through the tunnel should not. I did use OpenSSH for this before.

To be clear, I'm talking about the case where the http server closes the connection between the http server and the remote ssh server. And in such a case I think sshj should close the connection between the local end of the ssh tunnel and the http client.

hierynomus:
Can we add a test to prove that the tunnel is not broken :)

tsoiland:
> Can we add a test to prove that the tunnel is not broken :)

Yes, I'll have a look a it :)

hierynomus:
@tsoiland Any luck in adding a test? I want to try to push a new release out end of this week, would be good to have this in.

Issue #437, opened 2018-07-18
PKCS#8 support


jsmalkowski:
I'm debugging an application that uses sshj for connecting to other machines when given an ssh key pair. I uploaded a PKCS#8 encoded private key to this application, with which I verified I could manually authenticate to the other machines it manages through ssh (`ssh -i ~/.ssh/mykey.pem node`). However, the sshj library failed to authenticate with this key with the following stacktrace:

```
net.schmizz.sshj.userauth.UserAuthException: Exhausted available authentication methods
    at net.schmizz.sshj.SSHClient.auth(SSHClient.java:232)
    at net.schmizz.sshj.SSHClient.auth(SSHClient.java:208)
    at com.cloudera.server.cmf.node.NodeConfigurator.connect(NodeConfigurator.java:368)
    at com.cloudera.server.cmf.node.NodeConfigurator.configure(NodeConfigurator.java:965)
    at com.cloudera.server.cmf.node.NodeConfigurator.run(NodeConfigurator.java:1043)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.run(FutureTask.java:266)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at java.lang.Thread.run(Thread.java:748)
Caused by: net.schmizz.sshj.userauth.UserAuthException: Problem getting public key from PKCS8KeyFile{resource=[PrivateKeyStringResource]}
    at net.schmizz.sshj.userauth.method.KeyedAuthMethod.putPubKey(KeyedAuthMethod.java:46)
    at net.schmizz.sshj.userauth.method.AuthPublickey.buildReq(AuthPublickey.java:62)
    at net.schmizz.sshj.userauth.method.AuthPublickey.buildReq(AuthPublickey.java:81)
    at net.schmizz.sshj.userauth.method.AbstractAuthMethod.request(AbstractAuthMethod.java:63)
    at net.schmizz.sshj.userauth.UserAuthImpl.authenticate(UserAuthImpl.java:68)
    at net.schmizz.sshj.SSHClient.auth(SSHClient.java:226)
    ... 9 more
Caused by: java.io.IOException: Could not read key pair from: [PrivateKeyStringResource]
[...]

PR #708, opened 2021-08-16
Add PKCS8 parsing to support PEM ASN.1 Private Keys

exceptionfactory:
As described in issue #437, OpenSSH and other utilities support generating and storing private keys formatted using PKCS#8. The `PKCS8KeyFile` class uses the BouncyCastle `PEMParser`, which returns a `PrivateKeyInfo` object when reading a PEM file including the header `BEGIN PRIVATE KEY`. The `PKCS8KeyFile.readKeyPair()` method logs a debug message indicating that `PrivateKeyInfo` is not a supported object for parsing.

This pull request adds a `KeyPairConverter` interface with a primary implementation supporting `PrivateKeyInfo` objects returned from `PEMParser`. The `PrivateKeyInfoKeyPairConverter` delegates to specific implementations based on the Algorithm Object Identifier contained in `PrivateKeyInfo`. Implementations support reading DSA, ECDSA, and RSA Private Keys, and determining the associated Public Key in order to return a `PEMKeyPair`.

This pull request includes unit test updates with sample private keys generated using the following `ssh-keygen` commands:

- DSA: `ssh-keygen -t dsa -m PKCS8`
- ECDSA: `ssh-keygen -t ecdsa -m PKCS8`
- RSA: `ssh-keygen -t rsa -b 2048 -m PKCS8`

Classes included in this pull request should also provide the foundation for an additional implementation that supports reading encrypted private keys.


exceptionfactory:
Thanks for approving the initial workflow run @hierynomus. I pushed a commit adjusting the license header copyright year to match other headers and this passed running `./gradlew check`.

PR #716, opened 2021-09-01
Support v3 PuTTY keys

BerndSchuller:
Hi! 
With the latest PuTTY, the header line has changed to "PuTTY-User-Key-File-3".
This trivial patch will parse both old and new PuTTY key files correctly.

### DIFF
--- a/src/main/java/net/schmizz/sshj/userauth/keyprovider/PuTTYKeyFile.java
+++ b/src/main/java/net/schmizz/sshj/userauth/keyprovider/PuTTYKeyFile.java
@@ -45,6 +45,7 @@ import java.util.Map;
 
 /**
  * <h2>Sample PuTTY file format</h2>
+ *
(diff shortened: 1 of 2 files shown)


hierynomus:
Is that the only change? There is no format change involved?

Would it be possible to add a unit test with a new key type in there, so that we can verify this will keep on working.

BerndSchuller:
> Is that the only change? There is no format change involved?
> 
> Would it be possible to add a unit test with a new key type in there, so that we can verify this will keep on working.

yes, it's the only change, and yes I'll add a test

jpstotz:
This PR seems to enable support for unencrypted Putty v3 private keys. 

But the new file format also comes with a different encryption using a Argon2 based key derivation function. This is not supported by sshj therefore it should better throw an exception if the PPK key file contains an `Key-Derivation` entry.

Issue #706, opened 2021-08-12
BCrypt classpath hell


mindas:
I am experiencing a classpath hell which is related to how sshj incorporates bcrypt.

Consider sshj client code below:
```
        try (SSHClient sshClient = new SSHClient()) {
            sshClient.addHostKeyVerifier(new PromiscuousVerifier());
            KeyProvider keys = sshClient.loadKeys(settings.privateKey, settings.publicKey, PasswordUtils.createOneOff(settings.passphrase));
            sshClient.connect(settings.host, settings.port);
            sshClient.authPublickey(settings.username, keys);
```

This fails on the last line with the following exception:

```
java.lang.NoSuchMethodError: org.mindrot.jbcrypt.BCrypt.pbkdf([B[BI[B)V
 	at com.hierynomus.sshj.userauth.keyprovider.OpenSSHKeyV1KeyFile.initializeCipher(OpenSSHKeyV1KeyFile.java:150)
 	at com.hierynomus.sshj.userauth.keyprovider.OpenSSHKeyV1KeyFile.decryptBuffer(OpenSSHKeyV1KeyFile.java:132)
 	at com.hierynomus.sshj.userauth.keyprovider.OpenSSHKeyV1KeyFile.readDecodedKeyPair(OpenSSHKeyV1KeyFile.java:118)
 	at com.hierynomus.sshj.userauth.keyprovider.OpenSSHKeyV1KeyFile.readKeyPair(OpenSSHKeyV1KeyFile.java:85)
 	at net.schmizz.sshj.userauth.keyprovider.BaseFileKeyProvider.getPublic(BaseFileKeyProvider.java:81)
 	at net.schmizz.sshj.userauth.method.KeyedAuthMethod.putPubKey(KeyedAuthMethod.java:45)
 	at net.schmizz.sshj.userauth.method.AuthPublickey.buildReq(AuthPublickey.java:62)
 	at net.schmizz.sshj.userauth.method.AuthPublickey.buildReq(AuthPublickey.java:81)
 	at net.schmizz.sshj.userauth.method.AbstractAuthMethod.request(AbstractAuthMethod.java:68)
 	at net.schmizz.sshj.userauth.UserAuthImpl.authenticate(UserAuthImpl.java:73)
 	at net.schmizz.sshj.SSHClient.auth(SSHClient.java:221)
 	at net.schmizz.sshj.SSHClient.authPublickey(SSHClient.java:342)
[...]

PR #717, opened 2021-09-02
Move BCrypt class to avoid conflict with org.mindrot:jbcrypt

exceptionfactory:
As described in issue #706, the class name of `org.mindrot.jbcrypt.BCrypt` conflicts with a class of the same name in the original `org.mindrot:jbcrypt` library. With the SSHJ BCrypt class implementing the additional `pbkdf()` method, the two classes are not interchangeable. Relocating the `BCrypt` class under the `com.hierynomus.sshj` package structure avoids class path conflicts in applications that have dependencies on both SSHJ and JBcrypt libraries. Additional changes include renaming and updating the `BCryptTest` class to use standard JUnit Test annotations.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,7 +71,7 @@ license {
   mapping {
     java = 'SLASHSTAR_STYLE'
   }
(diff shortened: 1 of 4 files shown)


hierynomus (approved):
Makes sense!

Issue #678, opened 2021-04-07
Invalid module name: asn-one is not a Java identifier


avarsava:
Hello,

I am using sshj in a Java 14 application and I faced an issue getting my code to run.

```
java.lang.module.FindException: Unable to derive module descriptor for <project>/target/dependency/asn-one-0.5.0.jar
Caused by: java.lang.module.FindException: Automatic-Module-Name: com.hierynomus.asn-one: Invalid module name: 'asn-one' is not a Java identifier
```

I had to unpack the dependency jar, edit 'META-INF/MANIFEST.MF' to remove the '-' from the name, and repackage the jar to clear the error.

Thought you should know, and would appreciate any tips about working around this for future builds.

robertgates55:
Just hit the same thing!

PR #718, opened 2021-09-02
Replace com.hierynomus.asn1 with org.bouncycastle.asn1

exceptionfactory:
This pull request replaces uses of `com.hierynomus.asn1` classes with equivalent `org.bouncycastle.asn1` classes. With the direct dependency on Bouncy Castle for parsing private keys and other operations, this refactor removes the need for a dependency on `com.hierynomus:asn-one`. This addresses issue #678 and reduces the overall number of dependencies. Current unit tests for `SignatureDSA` and `SignatureECDSA` continue running as designed without modification.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,8 +49,6 @@ dependencies {
   implementation "org.bouncycastle:bcprov-jdk15on:$bouncycastleVersion"
   implementation "org.bouncycastle:bcpkix-jdk15on:$bouncycastleVersion"
   implementation "com.jcraft:jzlib:1.1.3"
-  implementation "com.hierynomus:asn-one:0.6.0"
-
   implementation "net.i2p.crypto:eddsa:0.3.0"
 
   testImplementation "junit:junit:4.12"
--- a/src/main/java/net/schmizz/sshj/signature/AbstractSignatureDSA.java
+++ b/src/main/java/net/schmizz/sshj/signature/AbstractSignatureDSA.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C)2009 - SSHJ Contributors
+ *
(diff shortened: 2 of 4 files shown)


exceptionfactory:
@hierynomus I rebased this PR from the current master branch. As mentioned in the description, since the Bouncy Castle library includes ASN.1 encoding and decoding, the refactored approach removes the need for the separate `asn-one` dependency.

exceptionfactory:
@hierynomus I noticed the following commit upgraded asn-one to 0.6.0 to address #678:

https://github.com/hierynomus/sshj/commit/7742d9b661c18571c6706db0137b69e4f8df83d4

If there are other reasons to avoid using the Bouncy Castle ASN.1 classes, feel free to close this PR.

hierynomus:
Hi, yes, I did choose to upgrade asn-one. Instead of adding more dependencies on BC, I've been striving to become more BC independent. Preferably it is possible to run SSHJ (in a limited mode) without having or depending on BC.

The reason for this, is that although BC is feature rich in encryption algorithms, it is for a lot of people hard to use the correct version of BC. Sometimes BC is already embedded in their `jre/lib/ext` folder with a specific version, or another component of their application depends on a specific version of BC. We've seen a huge number of issues in the past years with this. 

With https://github.com/hierynomus/smbj I've done the security provider abstraction better. I'm still thinking about doing the same for sshj.

So although I do value your contributions, this PR won't be merged. If you could send some PRs which do the inverse of this one, I would definitely welcome them!

exceptionfactory:
Thanks for the helpful reply @hierynomus, that makes sense. I will keep that in mind for future contributions, I appreciate the explanation!

Issue #642, opened 2020-10-20
Order proposed host key algorithms by known hosts


oakkitten:
Consider this scenario. We are connecting to a server with an RSA host key, and we store this key in our known hosts thing. The server then adds another key, say, Ed25519. Since sshj prefers Ed25519 over RSA, it will choose to verify using `ssh-ed25519`, which will obviously fail and will probably lead to a huge warning to boot.

OpenSSH [reorders](https://github.com/openssh/openssh-portable/blob/master@%7B2020-10-06T10:04:42Z%7D/sshconnect2.c#L122-L209) host keys algorithms by host key entries, putting the ones that can be actually used for verification at the front. It would be nice if sshj did the same.

While you can specify algorithms via configuration, it means that one have to create a new configuration for each server. Also, I don't think you can easily get available key types from `OpenSSHKnownHosts`.

I would expect something along `ssh.setAvailableHostKeyTypes(hostKeys.getAvailableHostKeyTypes(host, port))`

(Note that for this purpose it might be wise to consider ECDSA keys with different curves as different key types, which sshj seems to be doing anyway)

Issue #635, opened 2020-09-24
net.schmizz.sshj.transport.TransportException: Could not verify `ssh-ed25519` host key with fingerprint


gdufeZLYL:
net.schmizz.sshj.transport.TransportException: Could not verify `ssh-ed25519` host key with fingerprint 
        at net.schmizz.sshj.transport.KeyExchanger.verifyHost(KeyExchanger.java:210)
        at net.schmizz.sshj.transport.KeyExchanger.handle(KeyExchanger.java:368)
        at net.schmizz.sshj.transport.TransportImpl.handle(TransportImpl.java:517)
        at net.schmizz.sshj.transport.Decoder.decodeMte(Decoder.java:159)
        at net.schmizz.sshj.transport.Decoder.decode(Decoder.java:79)
        at net.schmizz.sshj.transport.Decoder.received(Decoder.java:231)
        at net.schmizz.sshj.transport.Reader.run(Reader.java:60)
===2020-09-24 20:36:22.566 [reader] INFO  net.schmizz.sshj.transport.TransportImpl Line:205 - Disconnected - HOST_KEY_NOT_VERIFIABLE

it's wrong, how to solve, help me, thanks

stjava:
I've got this problem, too

sysmat:
Me to using [sshj-0.31.0.jar:0.31.0], but in `cmd` I could connect with **ssh** and  **known_hosts**  to remote server

hpoettker:
Without more information like logs or code for a minimal example, it's very hard to give precise help. But I'll try. :smile: 

If host verification fails but you don't actually care about host verification, then you can turn it off by adding
`sshClient.addHostKeyVerifier(new PromiscuousVerifier())`
before calling `SSHClient::connect`.

If you care about host verification and want to use your `known_hosts`, make sure to call `SSHClient::loadKnownHosts` before connecting.

If you can connect with ssh but get the given exception with sshj, the problem is most likely that ssh and sshj negotiate different key algorithms with the remote server. Try connecting with `ssh -o HostKeyAlgorithms=ssh-ed25519 user@host`. If ssh warns you about conflicting host keys, remove the referenced line from `known_hosts` and try again. If ssh warns you that the authenticity of the remote host can't be established, proceed anyways and ssh will add the `ssh-ed25519` key to your `known_hosts`. Afterwards, you should be able to connect with sshj without disabling host verification.

[...]

PR #721, opened 2021-09-22
Prefer known algorithm for known host

WorkDayHeyHey:
(#642, #635... 10? issues)

Try to find the Algorithm that was used when a known_host
entry was created and make that the first choice for the
current connection attempt.

If the current connection algorithm matches the
algorithm used when the known_host entry was created
we can get a fair verification.

### DIFF
--- a/src/main/java/net/schmizz/sshj/transport/KeyExchanger.java
+++ b/src/main/java/net/schmizz/sshj/transport/KeyExchanger.java
@@ -170,20 +170,20 @@ final class KeyExchanger
     private void sendKexInit()
             throws TransportException {
         log.debug("Sending SSH_MSG_KEXINIT");
-        String knowHostAlg = findKnowHostAlg(transport.getRemoteHost(), transport.getRemotePort());
-        clientProposal = new Proposal(transport.getConfig(), knowHostAlg);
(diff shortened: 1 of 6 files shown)


WorkDayHeyHey:
I just make the changes again in a clean clone

hierynomus (approved):

hierynomus:
I've changed the implementation a bit to take into account all matching hostnames in the known hosts file, instead of only the first one. It is possible to have multiple keys for the same hostname. They're now taken into account in the order that was configured in the SshConfig object.

WorkDayHeyHey:
Nice! There is  so much complexity in this endeavor! I'm glad I could do something useful with so little specific knowledge.

I have used Ganymede for decades, but it hasn't been maintained for a decade. I have tried other libraries too, but if there is a problem most of us are out of our depth immediately. And I guess the knowledge gap is so wide people with knowledge don't have time to provide as much support as would be needed. So having something that works is great. Long live SSHJ!

Maybe I will contribute some examples. I am building a file system tool to allow my applications to access local and remote files without having to consider which they are. I think I have all the remote methods covered and working using SSHJ. I write in Scala, but I will make some examples up in Java, that will save you and others time.

Thanks!

hierynomus:
Good luck. We've done something like that once, have a look at https://github.com/xebialabs/overthere and https://github.com/xebialabs/nio-overthere

PR #724, opened 2021-09-24
Add parameter to limit read ahead to maximum length. Allows to use mu…

dkocher:
…ltiple concurrent threads reading from the same file with an offset without reading too much ahead for a single segment.


hierynomus:
@dkocher Is it possible to add a test for this?

dkocher:
@hierynomus Test has been added.

hierynomus (approved):

hierynomus:
Thanks, merged!

PR #729, opened 2021-10-02
only process supported Putty v3 keys + minor optimizations

jpstotz:
As mentioned in https://github.com/hierynomus/sshj/pull/716 the current Puttyv3 support is limited to unencrypted key files as the new file format uses Argon based key derivation functions . This PR adds checks so that we do not try to decrypt a key we can never decrypt successfully because of the missing Argon implementation.
Instead a describing exception is thrown so it is clear that the password was not the problem. 
A unit test for an encrypted RSA putty v3 key has been added.

Additionally I made some small optimizations on the putty key loading code.


hierynomus (approved):

hierynomus:
Thanks!

PR #730, opened 2021-10-03
full support for encrypted PuTTY v3 files (Argon2 library not included)

jpstotz:
Based on my patch yesterday I have implemented full support for encrypted PuTTY v3 keys.

As sshj is used on multiple platforms I wasn't sure about which Argon library to use. 
So in the end I just provided some interfaces that allows to use arbitrary Argon2 libraries but just providing two interface implementations (Argon2 libraries usually implement Argon2 in a native library and use JNI to call it).

The following example code uses `testImplementation group: 'de.mkammerer', name: 'argon2-jvm', version: '2.11'`
and shows how to to that.

And comments and critics on the selected interface design are welcome. The chosen implementation would allow different key derivation functions to implement. May be it is a bit over-engineered - @hierynomus would you prefer a specialized interface for Argon2?

```
    @Test
    public void testRSAv3EncryptedKey() throws Exception {
        PuTTYKeyFile key = new PuTTYKeyFile();
        key.init(new StringReader(v3_rsa_encrypted), new PasswordFinder() {
            @Override
            public char[] reqPassword(Resource<?> resource) {
                return "changeit".toCharArray();
            }

            @Override
            public boolean shouldRetry(Resource<?> resource) {
                return false;
            }
        });
        PuTTYKeyDerivationFactory factory = new PuTTYKeyDerivationFactory() {
            @Override
            public PuTTYKeyDerivation getKeyDerivation(final String keyDerivationMethod) {
                if (!keyDerivationMethod.matches("^Argon2(id|i|)$")) {
                    return null;
                }
                return new PuTTYKeyDerivation() {
                    @Override
                    public byte[] deriveKey(char[] passphrase, Map<String, String> headers, Map<String, String> payload) throws IOException {
                        Argon2Types at;
                        String algo = keyDerivationMethod.toLowerCase();
                        if ("argon2i".equals(algo)) {
                            at = Argon2Types.ARGON2i;
                        } else if ("argon2id".equals(algo)) {
                            at = Argon2Types.ARGON2id;
                        } else {
                            at = Argon2Types.ARGON2d;
                        }
                        Argon2 argon2 = Argon2Factory.create(at);
                        try {
                            // uses de.mkammerer:argon2-jvm unfortunately it does not expose the required methods,
                            // so we have to use reflection to get what we need
                            Method m = argon2.getClass().getDeclaredMethod("callLibraryRawHash",
                                    byte[].class, byte[].class, JnaUint32.class, JnaUint32.class,
                                    JnaUint32.class, byte[].class);
                            m.setAccessible(true);
                            byte[] hash = new byte[80]; // 32+16+32
                            byte[] pwd = new String(passphrase).getBytes();
                            byte[] salt = Hex.decodeHex(headers.get("Argon2-Salt"));
                            JnaUint32 iterations = new JnaUint32(Integer.parseInt(headers.get("Argon2-Passes")));
                            JnaUint32 memory = new JnaUint32(Integer.parseInt(headers.get("Argon2-Memory")));
                            JnaUint32 parallelism = new JnaUint32(Integer.parseInt(headers.get("Argon2-Parallelism")));
                            m.invoke(argon2, pwd, salt, iterations, memory, parallelism, hash);
                            return hash;
                        } catch (Exception e) {
                            throw new IOException(e);
                        }
                    }
                };
            }
        };
        key.addKeyDerivationFactory(factory);
        PrivateKey privateKey = key.getPrivate();
        assertNotNull(key.getPrivate());
        assertNotNull(key.getPublic());
    } 
```
Using this code the Puttyv3 key from unit test can be loaded including Hmac verification.

### DIFF
--- a/src/main/java/net/schmizz/sshj/userauth/keyprovider/PuTTYKeyFile.java
+++ b/src/main/java/net/schmizz/sshj/userauth/keyprovider/PuTTYKeyFile.java
@@ -38,9 +38,14 @@ import javax.crypto.spec.IvParameterSpec;
 import javax.crypto.spec.SecretKeySpec;
 import java.io.*;
 import java.math.BigInteger;
+import java.nio.CharBuffer;
+import java.nio.charset.StandardCharsets;
 import java.security.*;
 import java.security.spec.*;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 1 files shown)


jpstotz:
@hierynomus  Would be acceptable to add the unit test shown in my first comment?

exceptionfactory:
@jpstotz SSHJ currently includes a dependency on Bouncy Castle, which includes an Argon2 implementation, what do you think about providing an implementation based on the Bouncy Castle class [Argon2BytesGenerator](https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/crypto/generators/Argon2BytesGenerator.java)?

@hierynomus has mentioned attempting to avoid unnecessary dependencies on Bouncy Castle, but perhaps this is worth including?

jpstotz:
@exceptionfactory Thanks, I did not know that BouncyCastle already contains an Argon2 implementation. That really simplifies everything.

jpstotz:
I think this PR is finally complete and ready to merge. 

The unit tests now cover all the variants and they verify that the keys has been loaded correctly.

PR #733, opened 2021-10-13
Fix: if the client knows CA key, it should send host key algo proposal for certificates

vladimirlagunov:
My bad, I should have verified the new version before insisting on the release.

After updating the library, the tests written in advance on our side kept failing. Bisecting our code base didn't bring any results. Then I bisected SSHJ and found the commit where host key certificates broke down: 14bf93e677942ced047e7b6858d39f4376ee17dd The client started sending wrong host key algorithms proposals to the server for certificates. Undoubtedly, it's the tests problem that the host certificate tests didn't catch the bug.

Then I started figuring out why our local tests catched the bug, but tests from SSHJ didn't. Comparing good and bad OpenSSH server logs, I came across the way how kex proposals are generated. By default, Proposal.java sends an intersection of key types from the config, and key types from the known hosts file. However, if the intersection is empty, it sends the key types from the config. The intersection wasn't empty in the intellij tests, but was empty here, and the expected key type was sent to the server in these tests, not verifying the kex proposal logic. IntelliJ SSH tests use testcontainers, and are able to launch various servers. In contrast, these tests connect to a container started preliminary, therefore can't change the server configuration on the fly. That's why I had decided to write a filter for the host keys in the SSHJ test, which turned out to be buggy.

---

Luckily, the bug has an easy workaround on the users' side. I managed to get it working correctly on our side by overriding `findExistingAlgorithms` and adding a bit of reflection magic. I guess, everyone else could do the same for the first time. So, I don't treat the pull request as a show-stopper.

Indeed, the bug fix looks ugly. Feels like the knowledge about various key types is not encapsulated well. Please, propose a better fix if you know it.


hierynomus:
Would it make sense to use something like `testcontainers` for the integration testing, so that we can setup different containers for different scenarios. That would make this easier to test maybe?

vladimirlagunov:
Done.

This pull request also includes #741. It might be rebased on master when #741 is merged.

PR #741, opened 2021-11-10
Use testcontainers

vladimirlagunov:
It allows running different SSH servers with different configurations in tests, giving ability to cover more bugs, like mentioned in #733.


hierynomus:
🤩 Thanks @vladimirlagunov

PR #742, opened 2021-11-11
Lean on Config.keyAlgorithms choosing between rsa-sha2-* and ssh-rsa

vladimirlagunov:
Previously, there was a heuristic that was choosing rsa-sha2-512 after receiving a host key of type RSA. It didn't work well when a server doesn't have an RSA host key.

OpenSSH 8.8 introduced a breaking change: it removed ssh-rsa from the default list of supported public key signature algorithms. SSHJ was unable to connect to OpenSSH 8.8 server if the server has an EcDSA or Ed25519 host key.

Current behaviour behaves the same as OpenSSH 8.8 client does. SSHJ doesn't try to determine rsa-sha2-* support on the fly. Instead, it looks only on `Config.getKeyAlgorithms()`, which may or may not contain ssh-rsa and rsa-sha2-* in any order.

Sorry, this commit mostly reverts changes from #607.

Apparently, I tried to find a way to determine if the server supports some key algorithm signature. I haven't found anything related in source code of OpenSSH. So, I decided to implement in SSHJ the same approach as OpenSSH client uses.

Also, if this patch is accepted, better to add a note to the changelog about a backward compatibility breakage. The default config stops working with old servers like Apache SSHD that doesn't support rsa-sha2-* signatures. It is still possible to connect to them, but an instance of DefaultConfig should be modified a bit: `KeyAlgorithms.SSHRSA` should have higher priority than other RSA-related algorithms, which can be done by calling `Config.setKeyAlgorithms`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,7 +53,7 @@ dependencies {
   testImplementation "org.apache.sshd:sshd-core:$sshdVersion"
   testImplementation "org.apache.sshd:sshd-sftp:$sshdVersion"
   testImplementation "org.apache.sshd:sshd-scp:$sshdVersion"
-  testRuntimeOnly "ch.qos.logback:logback-classic:1.2.6"
+  testImplementation "ch.qos.logback:logback-classic:1.2.6"
   testImplementation 'org.glassfish.grizzly:grizzly-http-server:2.4.4'
   testImplementation 'org.apache.httpcomponents:httpclient:4.5.9'
   testImplementation 'org.testcontainers:testcontainers:1.16.2'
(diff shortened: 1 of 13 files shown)


vladimirlagunov:
Is there anything that should be modified to have it merged?

hierynomus:
No, nothing was blocking it, it more just escaped my attention that it wasn't merged yet.

Come to think of your starting coimment on this MR, it might be a good idea to add a factory method on DefaultConfig that adds the SSHRSA algorithm above the rest. So that it would be easy to adapt to the change for people. What do you think?

vladimirlagunov:
Added `ConfigImpl.prioritizeSshRsaKeyAlgorithm`.

hierynomus (approved):

hierynomus:
👍 Thanks!

PR #746, opened 2021-11-22
ByteBuffer.array() must not be used as it does not take the real buff…

dkocher:
…er size into account and returns the whole buffer up to its capacity.


hierynomus (approved):

hierynomus:
Thanks!

Issue #738, opened 2021-11-03
Naming the reader thread


kz282:
I would like to ask for a cosmetic change: please allow naming the net.schmizz.sshj.transport.Reader thread.

I am facing some issues during ssh connections and it is hard to find out the internals of the connection as the internal logs contain no specific id of the connection. It would be easier to filter logs based on thread names if the connected device ip address and some other ids could be put into the reader thread name. It could be stored in the net.schmizz.sshj.Config implementations. A Config.getReaderThreadName() and setReaderThreadName(String name) would be nice.

21.11.03 09:35:10.520 INFO  TransportImpl   [reader]: Received SSH_MSG_DISCONNECT (reason=PROTOCOL_ERROR, msg=Too many authentication failures for testuser)
21.11.03 09:35:10.520 ERROR TransportImpl   [reader]: Dying because - Too many authentication failures for testuser
net.schmizz.sshj.transport.TransportException: Too many authentication failures for testuser
        at net.schmizz.sshj.transport.TransportImpl.gotDisconnect(TransportImpl.java:565) ~[sshj-0.31.0.jar:0.31.0]
        at net.schmizz.sshj.transport.TransportImpl.handle(TransportImpl.java:521) ~[sshj-0.31.0.jar:0.31.0]
        at net.schmizz.sshj.transport.Decoder.decode(Decoder.java:113) ~[sshj-0.31.0.jar:0.31.0]
        at net.schmizz.sshj.transport.Decoder.received(Decoder.java:203) ~[sshj-0.31.0.jar:0.31.0]
        at net.schmizz.sshj.transport.Reader.run(Reader.java:60) ~[sshj-0.31.0.jar:0.31.0]
21.11.03 09:35:10.520 INFO  TransportImpl   [reader]: Disconnected - PROTOCOL_ERROR
21.11.03 09:35:10.520 DEBUG KeyExchanger    [reader]: Got notified of net.schmizz.sshj.transport.TransportException: [PROTOCOL_ERROR] Too many authentication failures for testuser
21.11.03 09:35:10.520 DEBUG UserAuthImpl    [reader]: Notified of net.schmizz.sshj.transport.TransportException: [PROTOCOL_ERROR] Too many authentication failures for testuser
21.11.03 09:35:10.520 DEBUG TransportImpl   [reader]: Setting active service to null-service
21.11.03 09:35:10.521 DEBUG Promise         [reader]: Setting <<transport close>> to `SOME`
[...]

PR #753, opened 2021-12-21
Add Thread naming based on remote socket address

exceptionfactory:
This pull requests adds standard Thread names for the Transport `Reader`, the SFTP `PacketReader`, and the Connection `KeepAlive` Threads.  The implementation includes a `ThreadNameProvider` class to set the Thread name prefixed with `sshj` followed by the simple class name and the remote socket address.  The following provides an example name for the Transport `Reader` Thread:

```
sshj-Reader-localhost/127.0.0.1:22
```
This approach to thread naming provides association with SSHJ, identification of the Thread implementation, and the ability to distinguish between connections to different endpoints. These changes address the troubleshooting challenges described in issue #738.


exceptionfactory:
Rebased

Issue #751, opened 2021-12-17
Update to support rename flags in #652 breaks V3 SFTP Protocol and causes errors with AWS SFTP Server


SteveReedMO:
The pull request for #652 enables the ability to set rename flags. This was recently rolled into V0.32.0.

I believe this is a SFTP V5+ only feature it should not be enabled for SFTP V3/4 servers that will not support this. 

As sshj has a max SFTP version of 3. It's rename packets should have the following structure

`       byte   SSH_FXP_RENAME`
 `       uint32     id`
 `       string     oldpath`
 `       string     newpath`

and not this structure, which is what V0.32.0 is currently sending for any rename

`      byte   SSH_FXP_RENAME`
`       uint32 id`
`       string oldpath`
`       string newpath`
`       uint32 flags`

Servers not expecting the extra uint32 flag at the end of the packet may experience issues.

We have witnessed the change in behaviour from V0.31.0 to V0.32.0 using this library with the AWS Transfer Family SFTP Server.

On a file rename we now see the following stacktrace

`net.schmizz.sshj.sftp.SFTPException: Timeout expired`
`at net.schmizz.sshj.sftp.SFTPException$1.chain(SFTPException.java:33)`
`at net.schmizz.sshj.sftp.SFTPException$1.chain(SFTPException.java:26)`
`at net.schmizz.concurrent.Promise.retrieve(Promise.java:139)`
`net.schmizz.sshj.sftp.SFTPEngine.doRequest(SFTPEngine.java:144)`
`net.schmizz.sshj.sftp.SFTPEngine.doRequest(SFTPEngine.java:144)`
`net.schmizz.sshj.sftp.SFTPEngine.remove(SFTPEngine.java:211)`
`net.schmizz.sshj.sftp.SFTPClient.rm(SFTPClient.java:129)`

and this error

`ERROR [Timer-Driven Process Thread-10] net.schmizz.concurrent.Promise <<sftp / 13>> woke to: net.schmizz.sshj.sftp.SFTPException: EOF while reading packet`

Issue #750, opened 2021-12-13
AWS SFTP: EOF while reading packet with version 0.32.0


pimenas:
Hello,

After upgrading to 0.32.0 we get the following exception when trying to upload to AWS SFTP service:
```
EOF while reading packet
net.schmizz.sshj.sftp.SFTPException: EOF while reading packet
	at net.schmizz.sshj.sftp.PacketReader.readIntoBuffer(PacketReader.java:55)
	at net.schmizz.sshj.sftp.PacketReader.getPacketLength(PacketReader.java:60)
	at net.schmizz.sshj.sftp.PacketReader.readPacket(PacketReader.java:76)
	at net.schmizz.sshj.sft
p.PacketReader.run(PacketReader.java:88)
```
With other sftp servers e.g. OpenSSH this doesn't happen. With 0.31.0 no error happens as well.

Attached a debug log of when this issue is reproduced
[test.log](https://github.com/hierynomus/sshj/files/7702639/test.log)

PR #754, opened 2021-12-21
Avoid setting SFTP rename flags below version 5

exceptionfactory:
Pull request #652 introduced support for SFTP rename flags, which was released in SSHJ 0.32.0. The `SFTPEngine.rename()` method includes the rename flag mask without checking the SFTP operative version, causing communication issues described in #750 and #751.

The [SFTP Version 5 Section 6.5](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-05#section-6.5) specification added support for rename flags, but previous versions such as [SFTP Version 3](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-03#section-6.5) do not support rename flags.

This pull request changes the behavior and avoids appending the rename flag mask to the SFTP request packet for operative versions less than 5.

### DIFF
--- a/src/main/java/net/schmizz/sshj/sftp/SFTPEngine.java
+++ b/src/main/java/net/schmizz/sshj/sftp/SFTPEngine.java
@@ -235,14 +235,17 @@ public class SFTPEngine
         if (operativeVersion < 1)
             throw new SFTPException("RENAME is not supported in SFTPv" + operativeVersion);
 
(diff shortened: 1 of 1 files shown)


hierynomus (approved):

hierynomus:
And thanks again! Totally missed that during the merge.

PR #756, opened 2021-12-23
Remove deprecated proxy connect methods from SocketClient

exceptionfactory:
This pull request removes deprecated `connect()` methods from `SocketClient` that support connections through a proxy server. As described in comments, proxy support can be implemented using a custom implementation of the standard Java `SocketFactory`.

The deprecated method comments indicated that these methods would be removed after version 0.12.0. Removing these deprecated methods provided the opportunity to remove the custom `Jdk7HttpProxySocket` class and the associated `JavaVersion` class.


hierynomus:
Actually not entirely correct, we still support Java7 for the rare enterprise case, so even for a SocketFactory you would need the custom Jdk7HttpProxySocket.

exceptionfactory:
Thanks for the feedback @hierynomus! What do you think about removing the connect methods, but retaining the Jdk7HttpProxySocket class?

exceptionfactory:
@hierynomus, following up on this, do you want to remove the deprecated `connect()` methods that accept `Proxy` parameters? If not, I can close this pull request.

exceptionfactory:
Sounds good, thanks!

PR #758, opened 2022-01-03
Refactor Base64 usage with standard Java and Bouncy Castle

exceptionfactory:
This pull request removes the custom `Base64` class and replaces references with interface-based access to new `Base64Encoder` and `Base64Decoder` components. The new `Base64Provider` class uses runtime reflection to determine the presence of `java.util.Base64`, using standard implementation classes when found. To provide compatibility with platforms that do not include `java.util.Base64` such as Java 7, the `Base64Provider` returns implementation classes based on the Bouncy Castle `Base64` class.

With Bouncy Castle being a required dependency for reading various key algorithms, this approach maintains compile and runtime compatibility with Java 7. Refactoring Base64 usage to interfaces also provides the opportunity to standardize and simplify the Base64 implementation in the event that Java 7 support is no longer needed.


hierynomus:
I'm not sure what the added benefit is of this. We have a well-tested public domain class in the codebase. And now we need to resort to magic to determine the right Base64 implementation. I'm actually not inclined to accept to PR

exceptionfactory:
> I'm not sure what the added benefit is of this. We have a well-tested public domain class in the codebase. And now we need to resort to magic to determine the right Base64 implementation. I'm actually not inclined to accept to PR

Thanks for the feedback @hierynomus! The initial idea was to replace the existing Base64 class with the `java.util.Base64` class in Java 8, but the requirement to continue supporting Java 7 required the more complicated implementation and fallback to Bouncy Castle.  With this approach, the Bouncy Castle fallback could be removed when Java 8 eventually becomes the minimum version.

Acknowledging that this trades removing the existing Base64 implementation for some amount of complexity, I will close this pull request. If and when Java 8 becomes the required minimum version, perhaps that would be an opportunity to replace the current approach with the Java 8 implementation.

PR #763, opened 2022-01-06
Try all public key algorithms available for a specific key type in SSH_MSG_USERAUTH_REQUEST

### LINKED ISSUE #761: Without support for `rsa-sha2-*` a connection is not possible anymore
We have updated to the latest version of sshj in Cyberduck and now there are reports that pubkey auth with RSA keys does not work anymore, e.g. https://github.com/iterate-ch/cyberduck/issues/12733.

This is caused by the removed heuristic based on the host key algorithms sent by the server. For older servers without support for `rsa-sha2-*` a connection is not possible anymore as in `SSH_MSG_USERAUTH_REQUEST` the pubkey algorithm `rsa-sha2-512` is chosen due to its position in the config list. To invoke `#prioritizeSshRsaKeyAlgorithm` would resolve the issue but is not an option as we don't know in advance what kind and version of server the user connects to. This would break the client for new server without `ssh-rsa` support. A workaround could be to try with and without `#prioritizeSshRsaKeyAlgorithm` but IMO this should be handled by the library.

As we cannot rely on the SSH extension `SSH2_MSG_EXT_INFO/server-sig-algs` I would suggest to repeat the `SSH_MSG_USERAUTH_REQUEST` request with all possible pubkey algorithms for a specific key type, e.g. `ssh-rsa, rsa-sha2-256, rsa-sha2-512` for RSA keys. That's the way it's done in the MINA client btw. Refer to https://issues.apache.org/jira/browse/SSHD-1105.

_Originally posted by @ylangisc in https://github.com/hierynomus/sshj/issues/740#issuecomment-1005763964_

ylangisc:
Fix #761.

### DIFF
--- a/src/main/java/net/schmizz/sshj/transport/Transport.java
+++ b/src/main/java/net/schmizz/sshj/transport/Transport.java
@@ -28,6 +28,7 @@ import net.schmizz.sshj.transport.verification.HostKeyVerifier;
 
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 /** Transport layer of the SSH protocol. */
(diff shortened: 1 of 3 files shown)


dkocher:
@hierynomus It would be great to have this reviewed.

hierynomus (approved):

Issue #183, opened 2015-03-11
ReadAheadRemoteFileInputStream


bkarge:
sshj.sftp.RemoteFile.ReadAheadRemoteFileInputStream.read() leads to data overwrites and/or ArrayIndexOutOfBoundsException when len < res.recvLen.

The problem does not occur with regular read() because there, the size of the response is limited by the len argument.

With read ahead, the size of the response is bounded by a _previous_ len argument.

This bug is not trivial to fix, because it relies on the RemoteFile.checkReadResponse swiss army knife, so there is no easy way to buffer partially evaluated response packets.

hierynomus:
Hi bkarge,

Do you have a sample or testcase to reproduce (preferably) both behaviours? That would help tremendously in fixing it.

bkarge:
Sir!

You use an ordered request fifo with entries (r+0, l1), (r+1, l2), ...
When read(myBuf,off,lk) is called, you enqueue (r+k, lk) and dequeue r+0, l1
Then you call checkReadResponse, which effectively calls arraycopy(...myBuf, off, l1) because SFTP will return a block of size l1 for the next response (if no problem did occur, but that's still another story)

If l1 > lk, then you overwrite the end of my buffer.

bkarge:
Below is a demonstration of the behaviour

I have submitted a pull request...

```
import net.schmizz.sshj.SSHClient;
import net.schmizz.sshj.sftp.OpenMode;
import net.schmizz.sshj.sftp.ReadAheadRemoteFileInputStream;
import net.schmizz.sshj.sftp.RemoteFile;
import net.schmizz.sshj.sftp.SFTPEngine;
import net.schmizz.sshj.transport.verification.HostKeyVerifier;

import java.io.InputStream;
import java.security.PublicKey;
import java.util.Arrays;
import java.util.EnumSet;
import java.util.Random;

public class SftpReadAheadTest {
    public static void main(String[] args) {
        try {
            SSHClient ssh = new SSHClient();
            ssh.loadKnownHosts();
            ssh.addHostKeyVerifier(new HostKeyVerifier()
            {
                @Override
                public boolean verify( String hostname, int port, PublicKey key ) {
                    return true;
                }
            });
            ssh.connect("localhost");
            ssh.authPublickey(System.getProperty("user.name"));

            SFTPEngine sftp = new SFTPEngine( ssh ).init();

            RemoteFile rf;
[...]

PR #769, opened 2022-03-04
Fix ReadAheadRemoteFileInputStream not reading the whole file if a buffer is too big

vladimirlagunov:
If an instance of ReadAheadRemoteFileInputStream before this change is wrapped into a BufferedInputStream with a big buffer, the SSH client requests big packets from the server. It turned out that if the server had sent a response smaller than requested, the client wouldn't have adjusted to decreased window size, and would have read the file incorrectly.

This change detects cases when the server is not able to fulfil client's requests. Since this change, the client adjusts the maximum request length, sends new read-ahead requests, and starts to ignore all read-ahead requests sent earlier.

Just specifying some allegedly small constant buffer size wouldn't have helped in all possible cases. There is no way to explicitly get the maximum request length inside a client. All that limits differ from server to server. For instance, OpenSSH defines SFTP_MAX_MSG_LENGTH as 256 * 1024. Apache SSHD defines MAX_READDATA_PACKET_LENGTH as 63 * 1024, and it allows to redefine that size.

Interestingly, a similar issue #183 was fixed many years ago, but the bug was actually in the code introduced for that fix.

### DIFF
--- a/src/main/java/net/schmizz/sshj/sftp/RemoteFile.java
+++ b/src/main/java/net/schmizz/sshj/sftp/RemoteFile.java
@@ -220,16 +220,42 @@ public class RemoteFile
 
     public class ReadAheadRemoteFileInputStream
             extends InputStream {
+        private class UnconfirmedRead {
+            private final long offset;
+            private final Promise<Response, SFTPException> promise;
+            private final int length;
+
+            private UnconfirmedRead(long offset, int length, Promise<Response, SFTPException> promise) {
+                this.offset = offset;
+                this.length = length;
+                this.promise = promise;
+            }
+
+            UnconfirmedRead(long offset, int length) throws IOException {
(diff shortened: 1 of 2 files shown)


vladimirlagunov:
After writing this simple piece of code, I stumbled upon questions: why do I remove already received data chunks if the window is adjusted? Why not reuse them later? The answer came after few hours of hacking. Indeed, it would be cool to make something as complicated as an interval tree. However, if we don't remove the data chunks which can be used only after many other smaller requests, the size of the unconfirmed reads will be equal to `maxUnconfirmedReads - 1` for a long time, and the read-ahead algorigthm won't work as expected. Another option might be increasing of `maxUnconfirmedReads`, but such behaviour would be unfair and unexpected.

hierynomus (approved):

hierynomus:
Thanks for the PR @vladimirlagunov. High quality as always!

PR #770, opened 2022-03-05
Add support for JuiceSSH generated ed25519 keys

TranceLove:
Reported from https://github.com/TeamAmaze/AmazeFileManager/issues/2976, it was found the key uses `aes-128-cbc` which is currently not supported by sshj. This change adds support for it.

To enable support for this, also eliminated hardcoding byte array size for key and IV, as a result of `BCrypt.pbkdf()`.


hierynomus:
Thanks, changes look good. Consider it merged!

hierynomus (approved):

PR #772, opened 2022-03-08
Remove jzlib dependency

p120ph37:
The JRE already contains a gzip implementation in `java.util.zip.*`, so we should be able to use that rather than relying on the external jCraft library.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -95,7 +95,7 @@ If you need something that is not included, it shouldn't be too hard to add (do 
 http://ssh-comparison.quendi.de/comparison.html[SSH Implementation Comparison]
 
 == Dependencies
-Java 6+. http://www.slf4j.org/download.html[slf4j] is required. http://www.bouncycastle.org/java.html[bouncycastle] is highly recommended and required for using some of the crypto algorithms.
+Java 7+. http://www.slf4j.org/download.html[slf4j] is required. http://www.bouncycastle.org/java.html[bouncycastle] is highly recommended and required for using some of the crypto algorithms.
 
 == Reporting bugs
 Issue tracker: https://github.com/hierynomus/sshj/issues
--- a/README.adoc
+++ b/README.adoc
@@ -95,7 +95,7 @@ If you need something that is not included, it shouldn't be too hard to add (do 
 http://ssh-comparison.quendi.de/comparison.html[SSH Implementation Comparison]
 
 == Dependencies
-Java 6+. http://www.slf4j.org/download.html[slf4j] is required. http://www.bouncycastle.org/java.html[bouncycastle] is highly recommended and required for using some of the crypto algorithms. http://www.jcraft.com/jzlib/[jzlib] is required for using zlib compression.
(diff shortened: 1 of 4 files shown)


hierynomus:
Unfortunately I do not think this will work. Zip is different from zlib and both algorithms are unable to read eachother's data.

p120ph37:
Yes, PKZip is a different format from Gzip, and long ago there was no Gzip implementation in the JRE, but since Java 6 or so, the `java.util.zip` package contains some classes to support Gzip now too.  See the docs here relating to the specific classes I used, for example: https://docs.oracle.com/javase/8/docs/api/index.html?java/util/zip/package-summary.html
Note it says "compression using the popular ZLIB compression library".

Additionally, I have tested this myself against a vsftp server by doing this to force the use of my Zlib implementation (note the absence of the "NoneCompression.Factory"):
```
ssh.getTransport().getConfig().setCompressionFactories(Arrays.asList(
    new DelayedZlibCompression.Factory()
);
```

Give it a try -- it works for me 😁

p120ph37:
I dug into the history a little more - it actually looks like zlib deflate/inflate support has been available in `java.util.zip` since all the way back in Java 1.1, when the JAR format was first introduced!
https://www.cs.princeton.edu/courses/archive/fall97/cs461/jdkdocs/api/java.util.zip.Deflater.html

The only reason the jCraft jzlib project was even created was because the JRE implementation didn't yet support partial-flush mode, as explained here:
http://www.jcraft.com/jzlib/

This deficiency was eventually fixed in Java 7 with the introduction of the 4-argument `deflate(byte[] b, int length, int off, int flush)`, allowing the use of the `SYNC_FLUSH` flag:
https://docs.oracle.com/javase/7/docs/api/java/util/zip/Deflater.html#deflate(byte[],%20int,%20int,%20int)

The Java bug report about this is here (and confirms that it was fixed as of JDK 7):
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4206909

So I suppose if you are intending to still support Java 6 for now, maybe you can't merge this yet, but as soon as you are targeting Java 7 and above, this PR should be usable.

hpoettker:
Is there a chance that this PR moves forward? 
I think that would be great! Just like jsch, jzlib seems to be unmaintained. And if jzlib can even be replaced by JDK internals, I think it's a good idea to do so.

Apache Mina Sshd did the same a while ago: https://github.com/apache/mina-sshd/commit/ea72a9eefa9872f6d41bd6c60a9a929824328051

p120ph37:
It looks like Java 7 may have already been a requirement since this commit: https://github.com/hierynomus/sshj/commit/35266945585b440ca75040d0af07181598531139#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R88

So I'm not aware of any reason this shouldn't be mergeable at this point.  @hierynomus , do you have any remaining concerns?

hierynomus (approved):

PR #773, opened 2022-03-16
Throw IOE instead of NPE if OpenSSHKeyV1KeyFile reads an empty file

vladimirlagunov:
There is a contract that `FileKeyProvider.readKey` throws an IOException if something goes wrong. Throwing an NPE is not expected by API users. Also, it is much more difficult to find out if the NPE is thrown due to a broken key file, or due to an internal bug.


hierynomus (approved):
Thanks for fixing the bug ;)

Issue #784, opened 2022-05-03
KeepAlive example incorrect


zman0900:
It seems #752 was a breaking change for how KeepAlive must be enabled.  Previously, KeepAlive::setKeepAliveInterval would actually start the KeepAlive thread, so it could be (or had to be?) called after calling SSHClient::connect.  Now, it appears the thread is only started from SSHClient::onConnect, which is called during SSHClient::connect.  So if the keep alive interval is not pre-set before calling connect, the thread never starts.  

The `net.schmizz.sshj.examples.KeepAlive` class should be changed to show setting the interval before connecting.

jkosternl:
@zman0900 , please close this issue, as the PR is merged, which fixes the issue. Thanks!

PR #791, opened 2022-06-09
Update KeepAlive and RemotePF examples

exceptionfactory:
As requested in issue #784, this pull request updates the `KeepAlive` example class to set the KeepAlive interval prior to connecting, following the behavior adjustments introduced in SSHJ 0.33.0. The `RemotePF` example follows the same approach for the Heartbeat setting, and removes the call to `Transport.setHeartbeatInterval()`, which is no longer defined.


jkosternl:
Good addition @exceptionfactory. @hierynomus, can you merge this one in please? I had issues with the examples, as what this PR fixes.

PR #793, opened 2022-06-09
Replace PKCS5 Key File Class with PKCS8

exceptionfactory:
This pull request removes the `PKCS5KeyFile` class and references in favor of parsing using `PKCS8KeyFile`. The Bouncy Castle `PEMParser` already used in `PKCS8KeyFile` supports both types of key files.

[RFC 8018](https://datatracker.ietf.org/doc/html/rfc8018) describes PKCS5 as the Password-Based Cryptography Specification Version 2.0, including methods for key derivation using various algorithms. Although encrypted private keys make use of password-based cryptography, using PKCS5 as a type of key file does not describe the key format itself.

[RFC 8017](https://datatracker.ietf.org/doc/html/rfc8017) describes PKCS1 as RSA Cryptography Specifications Version 2.2. [Appendix A](https://datatracker.ietf.org/doc/html/rfc8017#appendix-A) of RFC 8017 describes the ASN.1 syntax for RSA Public and Private Keys. The encoded binary contents of files starting with `BEGIN RSA PRIVATE KEY` or `BEGIN RSA PUBLIC KEY` align with the ASN.1 described in RFC 8017. For this reason, it is more accurate to describe the format of these files as PKCS1.

[RFC 5958](https://www.rfc-editor.org/rfc/rfc5958) describes PKCS8 Asymmetric Key Packages, including the ASN.1 syntax for different types of keys.

Although the naming of `PKCS8KeyFile` is accurate, the implementation leverages the Bouncy Castle `PEMParser`, which is capable of handling a variety of PEM encoded formats, including both PKCS8 and PKCS1 keys, with or without encryption. Removing the `PKCS5KeyFile` implementation and defaulting parsing to `PKCS8KeyFile` removes the need for custom PEM parsing in `PKCS5KeyFile` while maintaining support for the same types of keys.

Renaming `PKCS8KeyFile` to something more generalized, such as `PEMKeyFile` might be more accurate, but the initial approach avoids renaming to minimize the impact of changes.

Java supports unencrypted PKCS8 processing using [PKCS8EncodedKeySpec](https://docs.oracle.com/javase/7/docs/api/java/security/spec/PKCS8EncodedKeySpec.html) and other classes. It might be possible to remove the dependency on Bouncy Castle with a custom PEM-parsing implementation, but it would be useful to streamline generalized PEM-encoded processing prior to considering those changes.


exceptionfactory:
Thanks for the feedback @hierynomus! I updated `PKCS8KeyFileTest` with a new test method to exercise retrying a password after initial failure.

hierynomus (approved):

Issue #810, opened 2022-09-14
Received SSH_MSG_UNIMPLEMENTED while exchanging keys with KeepAlive Enabled


exceptionfactory:
Starting several SSH connections across multiple threads can result in an SSH server sending `SSH_MSG_UNIMPLEMENTED` during key exchange as shown in the follow error log:

```
ERROR net.schmizz.sshj.transport.TransportImpl - Dying because - Received SSH_MSG_UNIMPLEMENTED while exchanging keys
net.schmizz.sshj.transport.TransportException: Received SSH_MSG_UNIMPLEMENTED while exchanging keys
	at net.schmizz.sshj.transport.TransportImpl.gotUnimplemented(TransportImpl.java:565)
	at net.schmizz.sshj.transport.TransportImpl.handle(TransportImpl.java:495)
	at net.schmizz.sshj.transport.Decoder.decode(Decoder.java:113)
	at net.schmizz.sshj.transport.Decoder.received(Decoder.java:200)
	at net.schmizz.sshj.transport.Reader.run(Reader.java:60)
```

This appears to occur more frequently in SSHJ 0.33.0 and 0.34.0 with SSH KeepAlive enabled. These versions include changes in pull request #752, which start the KeepAlive Thread in `SSHClient.onConnect()`.

The underlying problem appears to be related to the timing of sending and receiving `SSH_MSG_KEXINIT` packets. Sending an `SSH_MSG_KEXINIT` packet can be initiated from different locations. The `SSHClient.onConnect()` method calls `Transport.doKex()`, which calls `KeyExchanger.startKex()`. `KexExchange.handle()` can also call `KeyExchanger.startKex()` when receiving an `SSH_MSG_KEXINIT` packet. This can cause problems with multiple connections due to different threads calling `KeyExchanger.startKex()` at different times.

When an OpenSSH server receives an unexpected `SSH_MSG_KEXINIT` packet, it sends an `SSH_MSG_UNIMPLEMENTED` packet and logs the following error:

```
dispatch_protocol_error: type 20 seq 3 [preauth]
```

The [atmoz/sftp](https://hub.docker.com/r/atmoz/sftp) Docker image provides a standard OpenSSH server for testing, but this issue should be reproducible with any recent version of OpenSSH server.

[...]

PR #811, opened 2022-09-14
Add Transport.isKeyExchangeRequired() to avoid unnecessary KEXINIT

exceptionfactory:
This pull request addresses SSH_MSG_UNIMPLEMENTED errors during key exchange described in issue #810.

Changes include adding `isKeyExchangeRequired()` to the `Transport` interface, with an implementation in `TransportImpl` that checks the status of Key Exchange completion and processing.

Checking `KeyExchanger.isKexDone()` indicates that at least one key exchange has not been completed, and checking `KeyExchanger.isKexOngoing()` indicates that a key exchange is not in progress.  Checking `Transport.isKeyExchangeRequired()` in `SSHClient.onConnect()` avoids sending an additional `SSH_MSG_KEXINIT` packet when a key exchange is already in progress or completed.

Additional changes include adding a started timestamp when setting a thread name in `ThreadNameProvider`. The started timestamp makes it easier to distinguished between threads for multiple connections to the same remote host and port.

### DIFF
--- a/src/main/java/com/hierynomus/sshj/common/ThreadNameProvider.java
+++ b/src/main/java/com/hierynomus/sshj/common/ThreadNameProvider.java
@@ -29,7 +29,8 @@ public class ThreadNameProvider {
     public static void setThreadName(final Thread thread, final RemoteAddressProvider remoteAddressProvider) {
         final InetSocketAddress remoteSocketAddress = remoteAddressProvider.getRemoteSocketAddress();
         final String address = remoteSocketAddress == null ? DISCONNECTED : remoteSocketAddress.toString();
-        final String threadName = String.format("sshj-%s-%s", thread.getClass().getSimpleName(), address);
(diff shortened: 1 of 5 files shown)


hierynomus (approved):

exceptionfactory:
Thanks for the feedback @hierynomus!

The test succeeded on a local build, but it appears to be an intermittent timing issue where the KeepAlive Thread may not have started. I pushed a new commit that moves the `TIMED_WAITING` status assertion after attempting authentication, which should make the test more reliable.

hierynomus (approved):

hierynomus:
Merged! Thanks for the improvement

exceptionfactory:
Thank you very much @hierynomus! Are you planning to do a new release of SSHJ soon? This would be very helpful for downstream consumers, and the internal nature of the Transport changes make it difficult to patch around without upgrading.

PR #813, opened 2022-09-15
Fix #805: Prevent CHANNEL_CLOSE to be sent between Channel.isOpen and…

### LINKED ISSUE #805: ieof packet referred to nonexistent channel
In my project, I have logic like this:
```java
  SSHClient sshClient = new SSHClient();
  sshClient.addHostKeyVerifier(new PromiscuousVerifier());
  sshClient.connect("host");
  sshClient.authPassword("xxx", "xxx");

  for (int i = 0; i < 100; i++) {
      try (Session sshSession = sshClient.startSession()) {
          Session.Command sshCommand = sshSession.exec("echo hello");
          IOUtils.readFully(sshCommand.getInputStream());
      }
  }

  sshClient.disconnect();
```

However, this throws IOException "ieof packet referred to nonexistent channel",  especially when connecting localhost and running in parallel **after I upgrade sshj from 0.31 to 0.32**.

kegelh:
… a Transport.write call

Otherwise, a disconnect with a "packet referred to nonexistent channel" message can occur.

This particularly happens when the transport.Reader thread passes an eof from the server to the ChannelInputStream, the reading library-user thread returns, and closes the channel at the same time as the transport.Reader thread receives the subsequent CHANNEL_CLOSE from the server.

The issue was reproducible and debug output showed that after a CHANNEL_CLOSE from the server, sshj sporadically sent the CHANNEL_CLOSE from a Channel.close call first and afterwards the CHANNEL_EOF from AbstractChannel.gotClose. 

This commit ensures,that the CHANNEL_EOF is not sent anymore if Channel.close sent out CHANNEL_CLOSE first.

The Transport.write call in DataBuffer.flush could produce a similar issue with a data packet, so it was placed within the openCloseLock, too.

### DIFF
--- a/src/itest/groovy/com/hierynomus/sshj/ManyChannelsSpec.groovy
+++ b/src/itest/groovy/com/hierynomus/sshj/ManyChannelsSpec.groovy
@@ -1,0 +1,74 @@
+/*
+ * Copyright (C)2009 - SSHJ Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


hierynomus (approved):

hierynomus:
Thanks for the find on how this happened!

PR #815, opened 2022-09-18
Support SSHClient.authPassword on FreeBSD

kegelh:
FreeBSD "keyboard-interactive" prompt is "Password for user@host:"

### DIFF
--- a/src/main/java/net/schmizz/sshj/userauth/method/PasswordResponseProvider.java
+++ b/src/main/java/net/schmizz/sshj/userauth/method/PasswordResponseProvider.java
@@ -27,7 +27,8 @@ import java.util.regex.Pattern;
 public class PasswordResponseProvider
         implements ChallengeResponseProvider {
 
-    public static final Pattern DEFAULT_PROMPT_PATTERN = Pattern.compile(".*[pP]assword:\\s?\\z", Pattern.DOTALL);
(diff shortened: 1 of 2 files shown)


vladimirlagunov:
However, [Linux-PAM sends the message `Password: `](https://github.com/linux-pam/linux-pam/blob/530c9f9e2d746e1d168c6b17863debda7664ac7c/modules/pam_userdb/pam_userdb.c#L53) . By the way, old ProFTPd versions from Debian used to localize PAM messages sent over SSH.

kegelh:
All previous prompts are still matched, so "Password: " from Linux continues to work. It would just match the default remote PAM prompt of FreeBSD, too. 

Everyone can initialize AuthKeyboardInteractive with a PasswordResponseProvider with their own pattern (this is what we currently do), so this change is not really necessary. 

I just thought it would be nice to share this so other users wouldn't have to find this out when trying to connect to FreeBSD. The "password" method is disabled by default on FreeBSD, so "keyboard-interactive" is used for password authentication.

hierynomus:
Yes indeed, currently it's already relatively simple to provide your own pattern. Now I'm not against making the pattern more lenient, but let's at least introduce some tests for the different prompts.

kegelh:
Indeed, I will add a test.

hierynomus (approved):

PR #820, opened 2022-09-23
Retry partial authentication with all remaining auth methods (Fixes #812)

### LINKED ISSUE #812: Whether mixed password authentication and public key authentication are supported
I have a project that requires both password and public key authentication, please tell me how to achieve it. Is there an example

hierynomus:


src/main/java/net/schmizz/sshj/SSHClient.java
```diff
@@ -218,13 +219,30 @@ public void auth(String username, Iterable<AuthMethod> methods)
...
             try {
-                if (auth.authenticate(username, (Service) conn, method, trans.getTimeoutMs()))
+                AuthResult result = auth.authenticate(username, (Service) conn, method, trans.getTimeoutMs());
+
+                if (result == AuthResult.SUCCESS) {
                     return;
+                } else if (result == AuthResult.PARTIAL) {
+                    // Put all remaining methods in the tried list, so that we can try them for the second round of authentication
+                    while (it.hasNext()) {
+                        tried.add(it.next());
```
vladimirlagunov, line 236: Suppose there are three keys loaded on the client: A, B, C. Also, suppose that the server knows only keys B and C, and requires both keys to be provided by the client. (`-o AuthenticationMethods=publickey,publickey`)

If I understand this code correctly, the authentication flow would work like this:
* SSHJ sends the public key A.
* The server responds "failure". 
* SSHJ adds the public key A to `tried`.
* SSHJ sends public key B.
* The server responds "partial". 
* SSHJ adds the last public key C to `tried` and invokes authentication recursively.
* SSHJ sends public key A again. It's definitely known that the server doesn't accept this key, because it already refused it.
* The server expectedly responds "failure".
* SSHJ sends the public key C.
* The server responds "success".

It becomes a severe problem for users who have dozens of keys, mostly coming from SSH agent. Also, it becomes a problem with paranoidal servers that has a small `MaxAuthTries`.

Instead, OpenSSH has a special workaround for public keys. It takes the keys from a round-robin collection, so definitely unacceptable keys are not re-sent again and again.

vladimirlagunov:
I can backport some of our changes with plenty of tests, though they're in Kotlin, which would require time for complete rewriting to Java.

jamespowenjr:
Sorry for taking so long to get to this. I tested this branch with nifi and unfortunately I still get the same error of "net.schmizz.sshj.userauth.UserAuthException: Exhausted available authentication methods" when the order is backwards from how the AuthMethods are added to the client.

PR #827, opened 2022-10-27
adds fallback to posix-rename@openssh.com extension if possible and c…

klelifo:
adds fallback to posix-rename@openssh.com extension if possible and communicates possible problems with flags to the developer

Tested with openssh-sftp-server-1:8.9p1-3 (Ubuntu 22.04)


src/main/java/net/schmizz/sshj/sftp/SFTPEngine.java
```diff
@@ -237,16 +237,71 @@ public void rename(String oldPath, String newPath, Set<RenameFlags> flags)
...
+
+            // finally: let the user know that the server does not support what was asked
+            else throw new SFTPException("RENAME-FLAGS are not supported in SFTPv" + operativeVersion + " and no " +
+                        "supported server extension could be found to achieve a similar result.");
         }
 
+        // build and send request
+        final Request request = newRequest(type);
+
+        if (serverExtension != null)
```
hierynomus, line 296: Can you add `{` around the if-statement... Everywhere where that's not done is a leftover ;)

klelifo: Sure, i just pushed a corrected version.
Thanks for your quick reply.

hierynomus:
Maybe only thing missing is some (unit) tests

klelifo:
I added some simple tests to cover sftp.rename(). 
Hope this improves things, although I did not add any advanced tests that simulate the behaviour of OpenSSH at the server side to trigger the use of the posix-rename@openssh.com extension. 
Not sure how to implement this...

hierynomus:
The integration tests run against an OpenSSH server, so that should be testable.

klelifo:
Sorry, I don't have much time at the moment.
Just wanted to let you know that I'm still interested in getting this merged.
Will provide the test cases as soon as I find the time.

hierynomus:
The integration tests have been simplified significantly I think. It should not be hard to add a few tests to cover this behaviour now.

klelifo:
I tried to make the tests as future-proof as possible by allowing to request a specific SFTP protocol version. This way things will not break/become non-functional when sshj will support SFTPv5 or greater in the future.
As of now (MAX_SUPPORTED_VERSION = 3) the implemented fallback for rename flags will always be active anyway - I think.

src/main/java/net/schmizz/sshj/sftp/SFTPEngine.java
```diff
@@ -81,7 +81,18 @@ public String canonicalize(String path)
 
     public SFTPEngine init()
             throws IOException {
-        transmit(new SFTPPacket<Request>(PacketType.INIT).putUInt32(MAX_SUPPORTED_VERSION));
+        return init(MAX_SUPPORTED_VERSION);
+    }
+
+    public SFTPEngine init(int requestedVersion)
```
hierynomus, line 87: Make this method package protected, so the test can call it, but it not becomes part of the contract.

hierynomus (changes requested):
One small nitpick to prevent too many externally visible methods, otherwise great work!

src/main/java/net/schmizz/sshj/sftp/SFTPEngine.java
klelifo: No problem :)
To make this change I had to create a new test class in the net.schmizz.sshj.sftp package. 
Hope this is ok.

hierynomus (approved):

hierynomus:
Merged, thanks for the PR!

klelifo:
You're welcome. Thanks for your patience!

Issue #782, opened 2022-04-28
add a configuration option to prevent BouncyCastle from registering at runtime so that sshj is compatible with GraalVM's native mode


insinfo:
add a configuration option to prevent BouncyCastle from registering at runtime so that sshj is compatible with GraalVM's native mode

```console
C:\MyJavaProjects\java_native> C:\MyJavaProjects\java_native\app\build\native\nativeBuild\Example.exe
Exception in thread "main" com.oracle.svm.core.jdk.UnsupportedFeatureError: Trying to verify a provider that was not registered at build time: BC version 1.69. All providers must be registered and verified in the Native Image builder. 
        at com.oracle.svm.core.util.VMError.unsupportedFeature(VMError.java:89)
        at javax.crypto.JceSecurity.getVerificationResult(JceSecurity.java:403)
        at javax.crypto.JceSecurity.getInstance(JceSecurity.java:140)
        at javax.crypto.KeyAgreement.getInstance(KeyAgreement.java:280)
        at net.schmizz.sshj.common.SecurityUtils.registerSecurityProvider(SecurityUtils.java:94)
        at net.schmizz.sshj.common.SecurityUtils.register(SecurityUtils.java:280)
        at net.schmizz.sshj.common.SecurityUtils.isBouncyCastleRegistered(SecurityUtils.java:251)
        at net.schmizz.sshj.DefaultConfig.<init>(DefaultConfig.java:78)
        at net.schmizz.sshj.SSHClient.<init>(SSHClient.java:131)
        at demo.SshjExample.init(SshjExample.java:22)
        at demo.Example.main(Example.java:25)

C:\MyJavaProjects\java_native>
```

[...]

PR #828, opened 2022-10-28
#782: Make runtime registration of bouncycastle skippable

Hayvon:
Adressing #782 by adding a system property check which makes the runtime registration of bouncycastle skippable.

### DIFF
--- a/src/main/java/net/schmizz/sshj/common/SecurityUtils.java
+++ b/src/main/java/net/schmizz/sshj/common/SecurityUtils.java
@@ -276,13 +276,13 @@ public class SecurityUtils {
 
     private static void register() {
         if (!registrationDone) {
-            if (securityProvider == null && (registerBouncyCastle == null || registerBouncyCastle)) {
+            if (System.getProperty("org.graalvm.nativeimage.imagecode") == null && securityProvider == null && (registerBouncyCastle == null || registerBouncyCastle)) {
                 registerSecurityProvider("org.bouncycastle.jce.provider.BouncyCastleProvider");
                 if (securityProvider == null && registerBouncyCastle == null) {
                     LOG.info("BouncyCastle not registered, using the default JCE provider");
                 } else if (securityProvider == null) {
-                    LOG.error("Failed to register BouncyCastle as the defaut JCE provider");
-                    throw new SSHRuntimeException("Failed to register BouncyCastle as the defaut JCE provider");
+                    LOG.error("Failed to register BouncyCastle as the default JCE provider");
+                    throw new SSHRuntimeException("Failed to register BouncyCastle as the default JCE provider");
                 }
             }
             registrationDone = true;


src/main/java/net/schmizz/sshj/common/SecurityUtils.java
```diff
@@ -275,17 +275,20 @@ public static synchronized void setSecurityProvider(String securityProvider) {
     }
 
     private static void register() {
-        if (!registrationDone) {
+        String noBCRegistrationAtRuntime = System.getProperty("bouncycastle.registration.runtime");
```
hierynomus, line 278: Instead of doing it like this, let's make it just a configuration in the config object. This will allow a user of the library to choose how to configure it, either through an env variable or through his own config.
Now we would be forcing them to set an env variable.

hierynomus (changes requested):

src/main/java/net/schmizz/sshj/common/SecurityUtils.java
Hayvon: Thanks for the quick feedback! 👍  Please take another look.

Hayvon: any updates? :)

src/main/java/net/schmizz/sshj/DefaultConfig.java
```diff
@@ -96,6 +98,19 @@ private String readVersionFromProperties() {
         }
     }
 
+    private boolean readBouncycastleRuntimeRegistrationProperty() {
+        try {
+            Properties properties = new Properties();
+            properties.load(DefaultConfig.class.getClassLoader().getResourceAsStream("sshj.properties"));
+            String property = properties.getProperty("sshj.bouncycastle.runtime.registration");
```
exceptionfactory, line 105: This approach requires reading a file from the class path, it seems like it would be better to avoid the lookup and make this a simple member variable of the `DefaultConfig` class.

Hayvon: Its already possible to set the variable programmatically by using the DefaultConfig.

exceptionfactory: Thanks for the clarification, that makes sense.

Hayvon: Thanks for the feedback. I actually misunderstood. I have taken out the property so that you can only set it via the ```DefaultConfig```. Any other feedback you have?

Hayvon: any updates? :)

src/main/java/net/schmizz/sshj/DefaultConfig.java
```diff
@@ -73,6 +73,7 @@
     public DefaultConfig() {
         setLoggerFactory(LoggerFactory.DEFAULT);
         setVersion(readVersionFromProperties());
+        SecurityUtils.setBouncycastleRuntimeRegistration(isBouncycastleRuntimeRegistrationEnabled());
```
hierynomus, line 76: If you would add some tests, you'd see that you can never ever change the configuration flag, as it is already passed down in the constructor of the config if you do it like this.

hierynomus (changes requested):

src/main/java/net/schmizz/sshj/DefaultConfig.java
Hayvon: Thanks, you are right. Will address it asap

Hayvon: any updates?

Hayvon:
The current approach disables the bouncycastle runtime registration using the defaultconfig, but it does not work for GraalVM Native Image because the defaultconfig is called too late at runtime.

Hayvon:
New approach to skip the bouncycastle registration at runtime is to use a system property set by GraalVM. This is also used by [Spring](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/NativeDetector.java) to determine if a native image is running.

hierynomus:
I will reiterate on what I've said: 

> I think you've misunderstood. It's up to the application that integrates sshj to decide how to pass the config to sshj. This kind of property or env reading code does not belong in the lib but rather in the surrounding app

exceptionfactory:
@Hayvon, I opened pull request #861 that attempts to address the Bouncy Castle registration issue. Registration can be disabled in the current version of SSHJ using the `SecurityUtils.setRegisterBouncyCastle()` method, but there are a couple other places that expect it to be registered, so the pull request includes some additional changes that should help support this use case more generally.

exceptionfactory:
@Hayvon and @hierynomus is this pull request still necessary? With the changes in PR #861 supporting SSHJ configuration with Bouncy Castle registration, the changes here no longer appear to be required.

hierynomus:
Indeed, I'll close this PR as the changes have been addressed in a better way in #861

PR #835, opened 2023-01-06
TimeoutException message improved

jpstotz:
A minor improvement to include the actual timeout value into the exception message. 
This can get handy in complex scenarios where it is not obvious what timeout is configured for the command that had failed.


hierynomus:
Thanks!

PR #838, opened 2023-02-08
Replace Curve25519 class with X25519 Key Agreement

exceptionfactory:
This pull request addresses an unresolved issue in the `Curve25519DH` implementation related to the use of Bouncy Castle for [Curve25519](https://en.wikipedia.org/wiki/Curve25519) Diffie-Hellman Key Agreement as defined in [RFC 8731](https://www.rfc-editor.org/rfc/rfc8731).

The existing implementation creates an instance of `ECParameterSpec` but does not use it for key agreement processing. The existing implementation also uses a ported implementation of the Curve25519 algorithm in the `djb.Curve25519` class.

This pull request refactors the `Curve25519DH` to use the standard Java Cryptography `KeyPairGenerator` and `KeyAgreement` interfaces, which are instantiated in the parent `DHBase` class using the standard `X25519` algorithm identifier.

The `KeyPairGenerator` implementation is based on the registered Security Provider, which is Bouncy Castle in normal operation. Java 11 introduced direct support for `X25519`, so this implementation approach provides greater flexibility for future versions of SSHJ using the standard Java Cryptography interfaces.

Changing to the standard interfaces removes the need for the ported `djb.Curve25519` implementation class, and also removes direct references to Bouncy Castle classes from the `Curve25519DH` class. These changes pass existing Key Exchange integration tests, and also include a unit test with basic key generation operations.


hierynomus:
This will require a few more changes, because now the Curve25519 is no longer compatible with Java7 which is the target compilation version. I'm fine with removing support for older java versions and making the minimum version java11 from now on, but then you will also need to update the build files are part of this PR.

exceptionfactory:
Thanks for the feedback @hierynomus!

To clarify the comment on Java 11, these changes work without issue on Java 8, so there is no need to make the minimum project version Java 11.

The net effect of these changes is to use the Bouncy Castle implementation of `KeyAgreement` and `KeyPairGenerator` using the `X25519`. The current approach to Key Exchange registration appears to require Bouncy Castle, but to clarify the original comment, this approach _could_ enable this Key Exchange method on Java 11 _without_ the need for Bouncy Castle down the road.

exceptionfactory:
On a related note, if you are interested in making Java 8 the minimum version in a separate PR, that might have some other benefits, but based on testing these changes with Java 8, that should not be required in this PR.

hierynomus (approved):

Issue #850, opened 2023-03-29
Receiving SSH_MSG_UNIMPLEMENTED occasionally during key exchange


rasantel:
I found in our application logs several instances of SSHJ connection errors due to receiving `SSH_MSG_UNIMPLEMENTED` from an OpenSSH server. The server logs show that, in each case, it receives an extra `KEXINIT` from the client, which causes to response with the unimplemented message.

The SSHJ version is 0.35.0. It has the fix for https://github.com/hierynomus/sshj/issues/810, which addresses this issue to a large extent, but not completely. Interestingly, the keep alive thread was not enabled in our application.

I don't have direct access to monitor the production system, but I am able to reproduce the issue locally with the help of debugger breakpoints (to force specific thread interleavings)  and an openssh 8.9p1-3ubuntu0.1 server running on the same machine. The test program is:

```
    try (SSHClient client = new SSHClient()) {
        client.addHostKeyVerifier(new PromiscuousVerifier());
        client.connect("localhost");
    }
```

This test will normally pass. But if you put a breakpoint at https://github.com/hierynomus/sshj/blob/v0.35.0/src/main/java/net/schmizz/sshj/SSHClient.java#L815 and wait for a second or two before resuming execution, or add a `Thread.sleep(1000)` before that line, the program will hit a `net.schmizz.sshj.transport.TransportException: Received SSH_MSG_UNIMPLEMENTED while exchanging keys`.

The reason why this happens is that, if the reader thread get a `KEXINIT` first and completes the key exchange before the main thread is ready to start its own key exchange, the main thread at `KeyExchanger.startKex` will think that it's ok to proceed because by then `kexOngoing` is back to `false`. Note that the check `trans.isKeyExchangeRequired()` earlier in `SSHClient.onConnect()` helps detect that the reader thread is already exchanging keys, but if it's called before the reader thread does that, it will return `true` and allow the main thread to proceed with the key exchange.

In other words, the sequence of events is:
1. `trans.isKeyExchangeRequired()` returns true in the main thread, so the main thread calls `doKex`.
2. The reader thread receives a `KEXINIT` and completes the key exchange with the server.
3. The main thread reaches `KeyExchanger.startKex` and initiates a new key exchange because `kexOngoing` is back to `false`.
4. The server responds with an `UNIMPLEMENTED` message,

[...]

PR #851, opened 2023-03-30
Fix race condition causing SSH_MSG_UNIMPLEMENTED occasionally during key exchange

rasantel:
The problem and diagnosis are documented in issue https://github.com/hierynomus/sshj/issues/850.

At `KeyExchanger.startKex`, the main thread should skip the key exchange if the reader thread is already doing an exchange (as per `KeyExchanger.kexOngoing`) or an exchange has already completed (as per `KeyExchanger.done`). Currently, the main thread checks this earlier in `SSHClient.onConnect`, but the result can change between then and `KeyExchanger.startKex`. 

Thankfully, `KeyExchanger.startKex` checks `kexOngoing` again and atomically "acquires" it (by checking it and switching it to `true`) but there is no new check for `done`, which by now might have been set by the reader or keepalive threads. If that has happened, the main thread will not know and it will proceed with its own key exchange, which will result in a `SSH_MSG_UNIMPLEMENTED` from the server.

Note that I found this to happen with the OpenSSH server with the default configuration provided by Ubuntu Linux 22.04, as well as the OpenSSH server used by the company where this issue was first encountered. The Apache sshd fixture in the unit test suite does not complain when receiving extra key exchanges under the same conditions. I'm no expert in the ssh protocol though; I don't know if this is a behavioral difference between the two servers or if it has to do with configuration differences.

The proposed fix is to check that `done` is false _after_ acquiring `kexOngoing`. By checking this while owning `kexOngoing`, there is no risk of `done` being set by another thread doing its own key exchange at the same time simply because only one thread can own `kexOngoing` at a time. If `done` was already set by some other thread before the main thread acquired `kexOngoing`, then the solution "releases" `kexOngoing` (sets it back to false) and avoids a new key exchange.

With this solution alone, however, additional key exchanges won't be possible again, but they must be allowed when it's time to re-key. My experiments with my own OpenSSH 8.9p1-3ubuntu0.1 server show that the server will reject with `SSH_MSG_UNIMPLEMENTED` any extra key exchange before authentication completes. After authentication, new key exchanges are allowed by it. Therefore, after `kexOngoing` is acquired, the solution adds a second check to see if authentication has completed.

If the approach looks good to reviewers, I'll add unit tests. Although the issue is not reproducible with the current sshd fixture, unit tests can show that duplicate key exchanges are avoided before authentication.


hierynomus:
Hi @rasantel, sorry for coming back somewhat later to this. A few additional unit tests would be great, then I think this is in a mergeable state! 

If you could add those, I'll include the PR in the next release.

hierynomus:
Looks good, consider it merged.

PR #852, opened 2023-03-31
Remove unused bcrypt password hashing methods

exceptionfactory:
This pull request removes unused bcrypt password hashing methods and associated unit tests.

The OpenSSH Private Key format uses a password-based key derivation function described in [bcrypt_pbkdf](https://man.openbsd.org/bcrypt_pbkdf.3) documentation, which does not depend on the bcrypt password hashing methods. The `BCrypt` class retains methods necessary for key derivation, but does not need other methods related to password hashing and custom Base64 encoding.


hierynomus:
Given that it's tested code, copied from a third party (which, if I remember correctly, does not provide a nice Maven dependency), I'd be tempted to keep the class intact. I don't think it hurts to have it in?  Or do you disagree?

exceptionfactory:
> Given that it's tested code, copied from a third party (which, if I remember correctly, does not provide a nice Maven dependency), I'd be tempted to keep the class intact. I don't think it hurts to have it in? Or do you disagree?

Thanks for the feedback @hierynomus.

Although the code is copied from an third party, I think it is better to retain only the elements that SSHJ actually uses. As evidenced by the methods removed, bcrypt key derivation is quite different from bcrypt password hashing. There is a well-maintained [bcrypt password hashing library](https://github.com/patrickfav/bcrypt) but it does not support bcrypt key derivation. From the perspective of only including code that actually applies to SSHJ, it seems like it would be better to remove the unused methods. On the other hand, if you think there is value is retaining the original source without any changes, I can see why that might be helpful for reference purposes. Less is more when it comes maintenance, and this improves some overall quality metrics, so that is another positive.

If you prefer to keep the original class unchanged, I can understand that approach.

hierynomus:
Hmm Makes sense, can't remember I actually came across the bcrypt lib before you mentioned. Agreed. let's clean it up. (Actually the overall code coverage dropped ;) )

PR #860, opened 2023-05-22
Upgrade to gradle 7.6.1 to make JDK 17+ work

kajtzu:
Please accept this trivial patch to upgrade Gradle 7.0 -> 7.6.1.

Compiling with JDK 17 (LTS) does not work with 7.0 but would require at least 7.3. 

I chose to upgrade to latest and greatest to allow compilation also on more recent JDKs 18, 19 according to this https://docs.gradle.org/current/userguide/compatibility.html


kajtzu:
Please LMK if you need a CLA or similar.

Issue #893, opened 2023-09-05
Broken pipe errors after upgrading to 0.36.0


dkocher:
After upgrading from 0.35.0 to 0.36.0 I see the following failures on the transport socket when sending write requests in integration tests.

```
2023-09-05 18:02:45,767 [sshj-Reader-localhost/127.0.0.1:2202-1693929765710] ERROR net.schmizz.sshj.transport.TransportImpl - Dying because - Connection reset
java.net.SocketException: Connection reset
	at sun.nio.ch.NioSocketImpl.implRead(NioSocketImpl.java:323) ~[?:?]
	at sun.nio.ch.NioSocketImpl.read(NioSocketImpl.java:350) ~[?:?]
	at sun.nio.ch.NioSocketImpl$1.read(NioSocketImpl.java:803) ~[?:?]
	at java.net.Socket$SocketInputStream.read(Socket.java:982) ~[?:?]
	at net.schmizz.sshj.transport.Reader.run(Reader.java:50) [sshj-0.36.0.jar:?]
```

hierynomus:
Ai, that does not sound well. Could you try adding an integration test to reproduce this in the project?

ericparton:
I'm also seeing this issue when transferring files with SCP, but only with certain files

maltesmann:
I just made a PR with an integrationtest reproducing this bug. It seems like using compression (more precisely the `net.schmizz.sshj.transport.compression.DelayedZlibCompression`) breaks the connection here. This is probably related to https://github.com/hierynomus/sshj/pull/772

hpoettker:
@maltesmann Thanks for the test!

@dkocher @ericparton Can you try whether the change in #901 resolves your issues?

hierynomus:
I'll keep this one open for a bit longer until @dkocher and/or @ericparton responded, but the PR has been merged already as it fixed the broken test ;)

dkocher:
> I'll keep this one open for a bit longer until @dkocher and/or @ericparton responded, but the PR has been merged already as it fixed the broken test ;)

🎉 I can no longer reproduce with 5d040dd4.

PR #900, opened 2023-10-10
add failing integrationtest for compressed uploads

maltesmann:
see https://github.com/hierynomus/sshj/issues/893


hierynomus:
Thanks for the test! #901 built further on top of your PR, and just got merged. I'll release a new version of the lib later!

PR #901, opened 2023-10-11
Fix zlib compression

hpoettker:
This PR addresses #893 and builds on top of #900, which provided a failing test.


hierynomus:
Wow, that's a fast fix, hadn't had the time yet to look at the previous PR that added the failing test. Thanks for finding the bug. I'm guessing this is due to the change to using standard JDK gzip instead of depending on jzlib.

hierynomus (approved):

PR #925, opened 2024-01-22
connected sockets can be passed to the library

martin-volf:
The client should be able to accept sockets that have been already connected to a remote server.


hierynomus (approved):

hierynomus (approved):

martin-volf:
one more attempt at actually improving test coverage

Issue #933, opened 2024-03-06
Failed to connect to higher version OpenSSL server after configuring KeepAliveInterval


2211898719:
`ssh.getConnection().getKeepAlive().setKeepAliveInterval(60);`

![image](https://github.com/hierynomus/sshj/assets/77563386/691683cd-0deb-4b7f-93ae-d1185b57fbe2)

`
net.schmizz.sshj.transport.TransportException: strict KEX violation: unexpected packet type 2 (seqnr 1)
	at net.schmizz.sshj.transport.TransportImpl.gotDisconnect(TransportImpl.java:548) ~[sshj-0.38.0.jar:na]
	at net.schmizz.sshj.transport.TransportImpl.handle(TransportImpl.java:500) ~[sshj-0.38.0.jar:na]
	at net.schmizz.sshj.transport.Decoder.decode(Decoder.java:113) ~[sshj-0.38.0.jar:na]
	at net.schmizz.sshj.transport.Decoder.received(Decoder.java:200) ~[sshj-0.38.0.jar:na]
	at net.schmizz.sshj.transport.Reader.run(Reader.java:60) ~[sshj-0.38.0.jar:na]
`

`		<dependency>
			<groupId>com.hierynomus</groupId>
			<artifactId>sshj</artifactId>
			<version>0.38.0</version>
		</dependency>
`

2211898719:
![image](https://github.com/hierynomus/sshj/assets/77563386/629dedc5-6c7d-47d8-9cc3-0104a71d5004)

derklaro:
I had the same issue. Fixed by using a custom KeepAlive sender that just waits until the key exchange is completed before starting the hearbeat.

2211898719:
How to customize it

PR #934, opened 2024-03-07
Don't send keep alive signals before kex is done

hpoettker:
The PR should address #933.


derklaro:
I think there is a difference between waiting for `conn.equals(conn.getTransport().getService())` and `isKexDone`: the first version also awaits the authentication process to complete, the new solution just waits for the key exchange. I guess it's better to just wait until the auth is done as well, before sending out keep alive messages 🤷

hpoettker:
I've added a reproducing test case to `StrictKeyExchangeTest` which uses a `KeepAlive` that sends heart beats instantantiouly after being started.

I've also relaxed the assertions in `StrictKeyExchangeTest` to prevent it from tripping on odd orders in the underlying concurrent processes.

The originally proposed changed should also work. But I've now made a simpler change, i.e. only starting the keep alive thread after the key exchange is done.

hierynomus:
Thanks for the PR!

PR #939, opened 2024-04-22
Corrected spelling of ResponseCode No 7

MikeTraceur:


hierynomus:
Hi Mike,

Thanks, but since the enum is public and exposed, this is a potentially breaking change and as such I'm not really willing to merge it.

jansohn:
IMHO not fixing this is the wrong decision. Postponing it for a major release due to a potentially breaking change, OK. But not fixing it at all?

hierynomus:
Why would you think so? It doesn't change functionality... It's a harmless typo in an enum constant name.

jansohn:
Well, it's clearly written wrong and it feels sloppy to me when I see such things in code. I mean correcting it might break some users code but it's so easy to fix anyway. But, your library, your decision of course.

PR #944, opened 2024-05-14
Fix null check for public key parameter.

### LINKED ISSUE #907: No initialization of public key depending on constructor usage of file key provider
We had an issue [^1] that we failed to initialize the public key required for connections using certificates for authentication. This is caused by the implementation of OpenSSHKeyV1KeyFile and OpenSSHKeyFile requiring to use a particular constructor to make sure an attempt is made to initialize the public key [^2] .

[^1]: iterate-ch/cyberduck#15183
[^2]: https://github.com/hierynomus/sshj/blob/ff4a4774bd1e3d45e9e796c2b17ef835e7e6d67f/src/main/java/net/schmizz/sshj/userauth/keyprovider/OpenSSHKeyFile.java#L88-L92

dkocher:
Addendum for #908 as discussed in [^1]. Resolves #907.

[^1]: https://github.com/hierynomus/sshj/pull/908/files#r1600044608

### DIFF
--- a/src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
+++ b/src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java
@@ -121,7 +121,7 @@ public class OpenSSHKeyV1KeyFile extends BaseFileKeyProvider {
 
     @Override
     public void init(String privateKey, String publicKey, PasswordFinder pwdf) {
-        if (pubKey != null) {
+        if (publicKey != null) {
             try {
                 initPubKey(new StringReader(publicKey));
             } catch (IOException e) {
@@ -133,7 +133,7 @@ public class OpenSSHKeyV1KeyFile extends BaseFileKeyProvider {
 
     @Override
     public void init(Reader privateKey, Reader publicKey, PasswordFinder pwdf) {
-        if (pubKey != null) {
+        if (publicKey != null) {
             try {
                 initPubKey(publicKey);
             } catch (IOException e) {


hierynomus:
Actually the null check is not for the parameter, but was once meant to only ensure that a subsequent call to `init`  would not re-initialize. Need to check what the potential problem is.

diamondo25:
This has been blocked for ages. Here is my dirty solution for at least 1 init() call:
```java
package com.hierynomus.sshj.userauth.keyprovider;

import lombok.val;

import java.security.PublicKey;

/**
 * This class fixes a bug with init() checking for pubKey != null, which causes publicKey (its argument) to never initialize...
 * <a href="https://github.com/hierynomus/sshj/pull/944">This Github pull request will be merged some day...</a>
 */
public class OpenSSHKeyV1KeyFileHacked extends OpenSSHKeyV1KeyFile {
    private class DummyPublicKey implements PublicKey {
        @Override
        public String getAlgorithm() {
            return "";
        }

        @Override
        public String getFormat() {
            return "";
        }

        @Override
        public byte[] getEncoded() {
            return new byte[0];
        }
    }
    
    @Override
    public void init(String privateKey, String publicKey) {
        // Make sure we only prepare loading the publicKey when its not null, as it'd otherwise throw a NullPointerException later.
        if (publicKey != null) {
            try {
                val field = OpenSSHKeyV1KeyFile.class.getDeclaredField("pubKey");
                field.setAccessible(true);
                field.set(this, new DummyPublicKey());
            } catch (Exception e) {
                throw new RuntimeException("Unable to hack pubKey field", e);
            }
        }
        
        super.init(privateKey, publicKey);
    }
}
````

hierynomus:
Superseeded by #1057

PR #951, opened 2024-07-11
Possibility to upload file using scp with minimum number of arguments

IgorMelnykSI:
It's possible that the remote host doesn't support certain options such as -p or -d, especially when it comes to a custom operating system. In this case, it's important to be able to disable any additional options used by default, which is unfortunately not possible at present.


hierynomus:
Fixed differently, thanks for the suggestion.

PR #971, opened 2024-12-28
ProxyCommand support (Issue 970)

knoxg:
This PR adds a new connectVia() method to SocketClient ( the superclass of SSHClient ), which takes an InputStream and OutputStream to use for communications.

When these streams are supplied by a java.lang.Process object, this gives sshj similar functionality to the 'ProxyCommand' directive in openssh, which enables the use of sshj with AWS SSM and EC2 instance connect.

An example class is also provided.


knoxg:
@hierynomus - wasn't sure why those checks weren't passing so tried pushing a small change to try to trigger it. 

From https://github.com/hierynomus/sshj/actions/runs/15792903957 it looks like this needs to be approved by a maintainer to run. Can you approve the workflow ?

KristijanL:
any updates on this?

hierynomus (approved):

Issue #973, opened 2024-12-31
Remove dependency on bc spec org.bouncycastle.jce.spec and use directly java.security.spec


uttamgupta:
If we remove dependency on bouncycastle spec and start using standard java security spec then it would help to support bc-fips 2.0 library. Other libraries like Jsch and Mina SSHD don't use bouncycastle spec so I can use bc-fips2.0 but I am not able to use Sshj due to this issue. 
Currently these three files used bouncycastle spec by importing ECNamedCurveSpec.
- src/main/java/net/schmizz/sshj/common/ECDSAVariationsAdapter.java
- src/main/java/net/schmizz/sshj/userauth/keyprovider/PuTTYKeyFile.java
- src/main/java/com/hierynomus/sshj/userauth/keyprovider/OpenSSHKeyV1KeyFile.java

exceptionfactory:
Thanks for summarizing the references to the Bouncy Castle ECNamedCurveSpec @uttamgupta.

I agree that moving more components in the direction of using the JCA would put SSHJ in a better position to use the BC-FIPS library.

There are several parts to making this happen, which should be considered individually.

For Edwards-Curve signatures, Java 15 added native support for Ed25519, so refactoring those references, with optional fallback to Bouncy Castle, could eliminate the current dependency on the `eddsa` library.

The PuTTYKeyFile class requires Bouncy Castle for the Argon2 key derivation function. Current Java versions do not provide native support for Argon2, so this is a hard dependency. However, perhaps some adjustments could make support for PuTTY encrypted keys optional based on the runtime availability of the required Bouncy Castle library.

The OpenSSH Key Version 1 implementation looks like it could be changed to use the standard Java Security spec components for ECDSA as suggested.

All that to say, this would be a multi-step effort requiring some evaluation based on supported Java versions, but worth it seems worth pursuing.

uttamgupta:
Thanks a lot @exceptionfactory for your response. Currently, my workflow relies solely on ECDSAVariationsAdapter.java. I’m curious about the level of effort required to add native support for Ed25519 and the specific steps needed to refactor this file alone. This would help unblock me and enable the use of bc-fips.

uttamgupta:
Hi @hierynomus I started working on this issue and created a pull request also - https://github.com/hierynomus/sshj/pull/976 .
It would be great if you could provide feedback on this issue.

hierynomus:
Both have been merged and are ready to be released

PR #984, opened 2025-02-18
Remove Bouncy Castle usage from Tests

exceptionfactory:
This pull request removes Bouncy Castle class references from tests.

Some changes involve removing unnecessary registration of the `BouncyCastleProvider` in several tests, and replacing the Bouncy Castle `Hex.toHexString()` with Apache SSHD `BufferUtils.toHex()` in other test cases.

This relates to a broader effort described in issue #973 for reducing usage of Bouncy Castle in general.


exceptionfactory:
@hierynomus This pull request is one of several, including #989 and #993 designed reduce or remove direct references to the Bouncy Castle library.

This particular change should be the most straightforward, since it is focused on changing tests as opposed to runtime components.

As you have availability to review, this would be useful for a number of integrations.

hierynomus (approved):

PR #987, opened 2025-02-21
Monitor keepalives with fewer threads

wylswz:


wylswz:
wait, com.hierynomus.sshj.KeepAliveTest this test need to be more deterministic

wylswz:
Not sure why it fails in workflow. Maybe that's related to cpu quota of the runner, I've relaxed some of the constraints (few connections, greater time skew) and see if it works.

Issue #929, opened 2024-02-15
SCP upload to Windows 11 Open SSH fails with no such file or directory error


robertpatrick:
We are using SSHJ to add remoting capabilities to our app.  We are currently testing cross-platform capabilities where the client is running on Linux connecting to a Windows 11 Enterprise server using the OpenSSH server provided by the Windows distribution.  SSH is working fine.  SCP download is working fine.  SCP upload is failing with:

```
Exception in thread "main" net.schmizz.sshj.xfer.scp.SCPRemoteException: Remote SCP command had error: scp: 'c:\Temp\model.yaml': No such file or directory
        at net.schmizz.sshj.xfer.scp.SCPEngine.check(SCPEngine.java:73)
        at net.schmizz.sshj.xfer.scp.SCPEngine.sendMessage(SCPEngine.java:133)
        at net.schmizz.sshj.xfer.scp.SCPUploadClient.sendFile(SCPUploadClient.java:105)
        at net.schmizz.sshj.xfer.scp.SCPUploadClient.process(SCPUploadClient.java:84)
        at net.schmizz.sshj.xfer.scp.SCPUploadClient.startCopy(SCPUploadClient.java:76)
        at net.schmizz.sshj.xfer.scp.SCPUploadClient.copy(SCPUploadClient.java:54)
        at net.schmizz.sshj.xfer.scp.SCPUploadClient.copy(SCPUploadClient.java:47)
        at net.schmizz.sshj.xfer.scp.SCPUploadClient.copy(SCPUploadClient.java:43)
        at net.schmizz.sshj.xfer.scp.SCPFileTransfer.upload(SCPFileTransfer.java:98)
        at net.schmizz.sshj.xfer.scp.SCPFileTransfer.upload(SCPFileTransfer.java:61)
        at net.schmizz.sshj.xfer.scp.SCPFileTransfer.upload(SCPFileTransfer.java:55)
        at com.oracle.test.SSHTest.main(SSHTest.java:36)
```

The odd thing is that scp from the command-line works fine.

```
[rpatrick@rpatrick-1 ssh-test]$ scp ./model.yaml rpatrick@<internal-hostname-redacted>:c:\\Temp\\model.yaml
model.yaml                                                                                                                                                                                100%  486   238.1KB/s   00:00
[rpatrick@rpatrick-1 ssh-test]$
```

[...]

PR #990, opened 2025-02-27
Update ScpCommandLine.java

BeauJBurroughs:
fix for SCP upload to Windows 11 Open SSH fails with no such file or directory error #929


hierynomus:
Might be worthwhile adding some tests to prove that this doesn't break things

robertpatrick:
are we waiting on something specific to review and merge this fix?  Seems like a simple fix for a critical issue...

hierynomus:
I'm wondering why this change? There are different Escape modes, have you tried the `DoubleQuote` escape mode... I don't see how the refenced issue is solved, there's no tests added that provide any explanatory code on why this works... So yes, it's waitring.

robertpatrick:
> I'm wondering why this change? There are different Escape modes, have you tried the `DoubleQuote` escape mode... I don't see how the refenced issue is solved, there's no tests added that provide any explanatory code on why this works... So yes, it's waitring.

I have a related issue open with lots of details but no one has bothered to offer a solution.

https://github.com/hierynomus/sshj/issues/929#issuecomment-1948969643

evilC:
Bump. Please fix!

> I'm wondering why this change? There are different Escape modes, have you tried the `DoubleQuote` escape mode

Is this something we can do in our code to work around the issue, or are you referring to how the PR should be coded?

robertpatrick:
@hierynomus one reason the current code doesn't work is that quoting command-line args with single quotes does not work in the Windows cmd shell (it does work in PowerShell).

```
C:\Temp>cd 'foo bar'
The system cannot find the path specified.

C:\Temp>cd "foo bar"

C:\Temp\foo bar>
```

robertpatrick:
I understand your comment about the `DoubleQuote` escape mode now.  The problem is that it requires some refactoring because the SCPFileTransfer.upload() method I am currently using doesn’t expose this option.  I will try it out but I shouldn’t be forced to use a different API to get the upload to work when the SSH server is running on Windows.  If this were my code, I would want it to “just work” by default.

robertpatrick:
@BeauJBurroughs @evilC I have followed the hints given by @hierynomus and closed #929.  Take a look at the code snippet at the bottom of that issue (https://github.com/hierynomus/sshj/issues/929#issuecomment-3517922522).

hierynomus:
Closed as the DoubleQuote escape does the trick

PR #999, opened 2025-03-18
Use try-with-resources

simon04:


src/main/java/net/schmizz/sshj/sftp/RandomAccessRemoteFile.java
```diff
@@ -318,10 +318,8 @@ public void writeShort(int v)
     public void writeUTF(String str)
             throws IOException {
         final DataOutputStream dos = new DataOutputStream(rf.new RemoteFileOutputStream(fp));
-        try {
+        try (dos) {
```
exceptionfactory, line 321: This changes is not supported on Java 8 according to the automated tests:

```
/home/runner/work/sshj/sshj/src/main/java/net/schmizz/sshj/sftp/RandomAccessRemoteFile.java:321: error: variables in try-with-resources are not supported in -source 8
        try (dos) {
                ^
  (use -source 9 or higher to enable variables in try-with-resources)
  ```
  
It looks like `fp` could be updated inside the try block.

exceptionfactory (changes requested):
Thanks for proposing the improvements @simon04.

The GitHub build workflow is failing due to an incompatibility with Java 8 as highlighted. Adjusting the approach should resolve the compatibility problem.

hierynomus:
Thanks for the PRs, keep them coming! 🎉

hierynomus (approved):

PR #1004, opened 2025-03-19
Convert utility classes to interface

simon04:
I like to define utility classes as interfaces. This clarifies that it does not hold any local state.

Feel free to dismiss thie PR if you dislike this style.


hierynomus:
Call me old-fashioned, but I think this is one where the Java guys made a boo-boo ;) Interfaces should not hold implementation.. Personally I dislike this pattern indeed.

exceptionfactory:
I agree with @hierynomus, keeping these utility methods in classes as opposed to interfaces seems like a better approach.

PR #1006, opened 2025-03-20
Set Bouncy Castle as optional feature in Gradle build

exceptionfactory:
This pull request assigns Bouncy Castle dependencies to the `bouncyCastle` [feature variant](https://docs.gradle.org/current/userguide/how_to_create_feature_variants_of_a_library.html) in the Gradle build. Assigning these dependencies to a variant results in marking the libraries as [optional dependencies](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html) in the published Maven configuration.

This change builds on recent efforts to move away from and isolate use of the Bouncy Castle library throughout the project.

This could be considered a breaking change for consumers that rely on inclusion of Bouncy Castle as a transitive dependency and require features that Bouncy Castle provides.

Direct features include reading PEM key files encrypted using historical methods other than PKCS8, and reading PuTTY Key Version 3 files that require the Argon2 key derivation function. Indirect features that depend on the Bouncy Castle security provider include support for Ed25519 when running on Java 14 or earlier, and support for X25519 when running on Java 10 or earlier. All of these features will continue to work when Bouncy Castle dependencies are declared directly.

The alternative to this change is retaining the current dependency requirements, and requiring consumers to exclude the Bouncy Castle transitive dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,11 +49,19 @@ configurations.implementation.transitive = false
 def bouncycastleVersion = "1.80"
 def sshdVersion = "2.15.0"
 
+java {
+    withJavadocJar()
+    withSourcesJar()
+    registerFeature('bouncyCastle') {
+        usingSourceSet(sourceSets.main)
+    }
+}
+
 dependencies {
   implementation "org.slf4j:slf4j-api:2.0.17"
-  implementation "org.bouncycastle:bcprov-jdk18on:$bouncycastleVersion"
-  implementation "org.bouncycastle:bcpkix-jdk18on:$bouncycastleVersion"
   implementation "com.hierynomus:asn-one:0.6.0"
+  bouncyCastleImplementation "org.bouncycastle:bcprov-jdk18on:$bouncycastleVersion"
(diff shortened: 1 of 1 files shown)


hierynomus:
I'll keep this one on the backlog for a moment. There's a ton of other changes pending, let's get them out first. This might warrant a slightly bigger version jump.

exceptionfactory:
Thanks @hierynomus, I agree that this change has a larger potential impact and could warrant a major version adjustment. Revisiting this later after other incremental releases sounds good!

exceptionfactory:
@hierynomus Do you have any additional thoughts on the feature variant approach proposed, making Bouncy Castle and optional dependency? If you would prefer to keep the current approach, I can close this one.

hierynomus:
No I like the approach here. Let's first release a new version with all the accumulated fixes and then this in a new version to isolate the potential breakage

exceptionfactory:
> No I like the approach here. Let's first release a new version with all the accumulated fixes and then this in a new version to isolate the potential breakage

Sounds like a plan, thanks for following up on this!

PR #1013, opened 2025-05-13
Ensure reset of security provider preference.

### LINKED ISSUE #1012: Bouncycastle registration in tests
The test `Curve25519DHTest` may break other tests as it alters security provider preference. Regression from f4f8071020bd4877f70dc9338d65819b77c8fbd0

dkocher:
Resolves #1012.


exceptionfactory (approved):
Thanks for finding and addressing this issue @dkocher, the updated approach in the test and resetting the register Bouncy Castle flag looks good.

PR #1023, opened 2025-09-26
Security: Prefer newer rsa versions (256 and 512) before legacy (ssh-rsa)

devlinx9:
When the host is added to known_hosts and it supports rsa-sha2-256 or rsa-sha2-512, the filterKnownHostKeyAlgorithms method put first ssh-rsa before the newer versions. That's because the signature is ssh-rsa in the known_hosts file. if the server only accepts as its first or only method rsa-sha2-512, the connection will fail since the preferred method on the client side is still the outdated ssh-rsa.


hierynomus:
Superseeeded by #1056

PR #1034, opened 2026-03-26
Make ChannelInputStream respect connection timeout

### LINKED ISSUE #731: SCP File uploads stuck if the remote didn't send status code
The file upload thread is stuck at SCPEngine.check method probably because the remote didn't respond with the completion status code. While these are possible due to network issues, the client might need to wait for some time and invoke a timeout with an error to the user. Please look at the thread dump

```
"TaskStateModelFactory-task_thread-6" #78 daemon prio=5 os_prio=0 cpu=39889.63ms elapsed=1726257.09s tid=0x00007f27b4073000 nid=0x23c6 in Object.wait()  [0x00007f27976f4000]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(java.base@11.0.9/Native Method)
	- waiting on <no object reference available>
	at java.lang.Object.wait(java.base@11.0.9/Object.java:328)
	at net.schmizz.sshj.connection.channel.ChannelInputStream.read(ChannelInputStream.java:103)
	- waiting to re-lock in wait() <0x0000000092ae8e60> (a net.schmizz.sshj.common.Buffer$PlainBuffer)
	at net.schmizz.sshj.connection.channel.ChannelInputStream.read(ChannelInputStream.java:86)
	- locked <0x0000000092ae8e78> (a [B)
	at net.schmizz.sshj.xfer.scp.SCPEngine.check(SCPEngine.java:60)
	at net.schmizz.sshj.xfer.scp.SCPUploadClient.sendFile(SCPUploadClient.java:100)
	at net.schmizz.sshj.xfer.scp.SCPUploadClient.process(SCPUploadClient.java:78)
	at net.schmizz.sshj.xfer.scp.SCPUploadClient.startCopy(SCPUploadClient.java:70)
	at net.schmizz.sshj.xfer.scp.SCPUploadClient.copy(SCPUploadClient.java:50)
	- eliminated <owner is scalar replaced> (a net.schmizz.sshj.xfer.scp.SCPUploadClient)	at net.schmizz.sshj.xfer.scp.SCPUploadClient.copy(SCPUploadClient.java:43)
	- eliminated <owner is scalar replaced> (a net.schmizz.sshj.xfer.scp.SCPUploadClient)	at net.schmizz.sshj.xfer.scp.SCPFileTransfer.upload(SCPFileTransfer.java:55)
```

jpstotz:
If the `ChannelInputStream` does not get new data it waits without timeout. 

This patch makes `ChannelInputStream` respect the timeout configured in the `Connection`.

fixes #731

### DIFF
--- a/src/main/java/net/schmizz/sshj/Config.java
+++ b/src/main/java/net/schmizz/sshj/Config.java
@@ -204,4 +204,12 @@ public interface Config {
     int getMaxCircularBufferSize();
 
     void setMaxCircularBufferSize(int maxCircularBufferSize);
+
+    int getChannelReadTimeoutMs();
+
+    void setChannelReadTimeoutMs(int channelReadTimeoutMs);
+
+    int getChannelErrorReadTimeoutMs();
+
+    void setChannelErrorReadTimeoutMs(int channelErrorReadTimeoutMs);
 }
--- a/src/main/java/net/schmizz/sshj/ConfigImpl.java
+++ b/src/main/java/net/schmizz/sshj/ConfigImpl.java
@@ -51,6 +51,8 @@ public class ConfigImpl
     private boolean verifyHostKeyCertificates = true;
     // HF-982: default to 16MB buffers.
     private int maxCircularBufferSize = 16 * 1024 * 1024;
+    private int channelReadTimeoutMs = 0;
+    private int channelErrorReadTimeoutMs = 0;
 
     @Override
     public List<Factory.Named<Cipher>> getCipherFactories() {
@@ -185,6 +187,26 @@ public class ConfigImpl
     @Override
     public void setMaxCircularBufferSize(int maxCircularBufferSize) {
         this.maxCircularBufferSize = maxCircularBufferSize;
+    }
+
+    @Override
+    public int getChannelReadTimeoutMs() {
+        return channelReadTimeoutMs;
+    }
+
+    @Override
+    public void setChannelReadTimeoutMs(int channelReadTimeoutMs) {
+        this.channelReadTimeoutMs = channelReadTimeoutMs;
+    }
+
+    @Override
+    public int getChannelErrorReadTimeoutMs() {
+        return channelErrorReadTimeoutMs;
+    }
+
+    @Override
+    public void setChannelErrorReadTimeoutMs(int channelErrorReadTimeoutMs) {
+        this.channelErrorReadTimeoutMs = channelErrorReadTimeoutMs;
     }
 
     @Override
--- a/src/main/java/net/schmizz/sshj/connection/channel/AbstractChannel.java
+++ b/src/main/java/net/schmizz/sshj/connection/channel/AbstractChannel.java
@@ -95,7 +95,7 @@ public abstract class AbstractChannel
         id = conn.nextID();
 
         lwin = new Window.Local(conn.getWindowSize(), conn.getMaxPacketSize(), loggerFactory);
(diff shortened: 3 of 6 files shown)


jpstotz:
In my first version I used [Object.wait(long)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Object.html#wait(long)) but this method has the drawback that you can not identify if it returns because of a `notifyAll()` call or if a timeout occurred. 

Therefore I changed the implementation from `synchronized(buf) {..}` blocks and `Object.wait(..)` `Object.notifyAll()` calls to `ReentrantLock` and `Condition` as the latter allows us to detect if there was a timeout or not when using 
[Condition.await​(long time, TimeUnit unit)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/locks/Condition.html#await(long,java.util.concurrent.TimeUnit))

peterjhill:
[P1] stderr stream ignores the timeout entirely.
SessionChannel.java (line 35) constructs err with timeout 0, so Session.getErrorStream().read(...) still blocks forever. That creates an inconsistent contract: stdout respects the connection timeout, stderr does not. For long-lived subsystem/command users, stderr can wedge a reader thread the same way stdout did.

[P2] Timeout is frozen when the channel is created.
AbstractChannel.java (line 98) passes conn.getTimeoutMs() as an int, and ChannelInputStream.java (line 44) stores it permanently. If callers change Connection#setTimeoutMs() after opening a channel, reads keep using the stale value. Other connection operations fetch conn.getTimeoutMs() at call time, so this is surprising.

[P2] Public constructor compatibility is broken.
ChannelInputStream.java (line 53) replaces the existing public 3-arg constructor with a 4-arg one. Even if this class is mostly internal by convention, it is public API in the jar. Keep the old constructor and delegate to a default timeout source.

[P3] No regression test for the timeout behavior.
The PR changes concurrency semantics in a transport stream but adds no test that a blocked read() times out, no test that EOF wins over timeout, and no stderr coverage. That is exactly how the err = timeout 0 bug slips through.

jpstotz:
@peterjhill Thank you for your detailed review. 

> [P2] Timeout is frozen when the channel is created.

Propagating time-out information from `Connection` respectively `ConnectionImpl` to `ChannelInputStream` will get very hard as both classes have no direct relation/reference to each other. The only way I see is adding something like a time-out changed listener system  `Connection` and `ConnectionImpl` and change `ChannelInputStream` so that it gets a reference to `Connection` where it can register for time-out changes. Is that what you had in mind or do you see a a more simply way?

Edit: An alternative would be saving the time-out in a MutableInteger alike instance (e.g. AtomicInteger) in `Connection` and expose it throug the API. That way changes to the time-out could be propagates automatically if you keep the reference and retrieve the actual value every time you need it.

> [P3] No regression test for the timeout behavior.

The main problem is that I don't really understand how the bug can be triggered. I just know that it happened in my application and I know that it happens because of some stream reading problem, but I am not familiar how `ChannelInputStream` fit's into the big picture of sshj. Therefore I have not provided unit tests.

hierynomus:
@peterjhill I think I've covered the missing parts. Keeping the timeout static once assigned is consistent with all other configurable settings in the Config.

pilengkyaw-sys (commented):
Leave10

peterjhill:
thank you!!
I help out with the juniper netconf-java library and use your library for all the ssh magic.
thanks for all the work you have put into sshj!

PR #1037, opened 2026-05-07
Add support for mlkem768x25519-sha256

### LINKED ISSUE #1017: Feature request: adding mlkem768x25519-sha256 post-quantum key exchange
Hello,

Are there any plans for SSHJ to support the new `mlkem768x25519-sha256` key exchange algorithm?

[OpenSSH 9.9](https://www.openssh.com/txt/release-9.9) (2024-09-19) introduced support for the new FIPS 203 Module-Lattice Key Encapsulation Mechanism (ML-KEM) post-quantum key exchange algorithm. ML-KEM (formally known as CRYSTALS Kyber) has been [formally standardized by NIST](https://csrc.nist.gov/pubs/fips/203/final) earlier last year, and is designed to be secure against quantum attacks. OpenSSH has chosen to support ML-KEM by using a PQ/T hybrid implementation: `mlkem768x25519-sha256`. Efforts are already underway standardising this via [an IETF Draft](https://datatracker.ietf.org/doc/draft-ietf-sshm-mlkem-hybrid-kex/).

[OpenSSH 10.0](https://www.openssh.com/txt/release-10.0) (2025-04-09) is now using this new quantum-resistant key exchange algorithm by default.

And, as recently announced, [OpenSSH 10.1](https://www.openssh.com/pq.html) will start warning if no post-quantum key exchange algorithm is supported by the server. This in an effort to speed-up adoption and help thwart "store now, decrypt later" attacks.

alexandrejbr:
Adds support for the mlkem768x25519-sha256 SSH key exchange, defined in draft-kampanakis-curdle-ssh-pq-ke and shipped by OpenSSH 9.9+ and Erlang/OTP 28.4+. It combines classical curve25519-sha256 with NIST FIPS-203 ML-KEM-768, providing post-quantum confidentiality (with the KEM provided by BouncyCastle) without giving up the security of X25519 if either primitive is later broken.

To keep compatibility with older versions I didn't take the JCA approach and instead created a bouncy castle wrapper which meant that I couldn't get for free things like the SecureRandom so one needs to be created explicitly. There is a Random factory but that is not guaranteed to be a SecureRandom as far as I understand.

EDIT: there's no direct dependency with bouncy castle in the updated version, but it's instead used as a fallback using reflection.

The changes of this PR were tested with openssh 10+ and Erlang OTP 28.4.1.

I kept groovy and spock-core upgrade in a separate commit as I am unsure that is wanted in this project but I needed it to test and may be nice to do anyway. The integration tests are as well in a separate commit as they require a different base image.

Fixes https://github.com/hierynomus/sshj/issues/1017

### DIFF
--- a/src/itest/java/com/hierynomus/sshj/SshdContainer.java
+++ b/src/itest/java/com/hierynomus/sshj/SshdContainer.java
@@ -103,9 +103,12 @@ public class SshdContainer extends GenericContainer<SshdContainer> {
     }
 
     public static class Builder implements Consumer<DockerfileBuilder> {
+        private static final String DEFAULT_BASE_IMAGE = "alpine:3.19.0";
+
         private List<String> hostKeys = new ArrayList<>();
         private List<String> certificates = new ArrayList<>();
         private @NotNull SshdConfigBuilder sshdConfig = SshdConfigBuilder.defaultBuilder();
+        private @NotNull String baseImage = DEFAULT_BASE_IMAGE;
 
         public static Builder defaultBuilder() {
             Builder b = new Builder();
@@ -116,6 +119,16 @@ public class SshdContainer extends GenericContainer<SshdContainer> {
 
         public @NotNull Builder withSshdConfig(@NotNull SshdConfigBuilder sshdConfig) {
             this.sshdConfig = sshdConfig;
+            return this;
+        }
+
+        /**
+         * Override the base image used to build the sshd container. Useful for tests that need
+         * a specific OpenSSH version (for example, OpenSSH&nbsp;≥10 for the
+         * {@code mlkem768x25519-sha256} key exchange).
+         */
+        public @NotNull Builder withBaseImage(@NotNull String baseImage) {
+            this.baseImage = baseImage;
             return this;
         }
 
@@ -148,7 +161,7 @@ public class SshdContainer extends GenericContainer<SshdContainer> {
 
         @Override
         public void accept(@NotNull DockerfileBuilder builder) {
-            builder.from("alpine:3.19.0");
+            builder.from(baseImage);
             builder.run("apk add --no-cache openssh");
             builder.expose(22);
             builder.copy("entrypoint.sh", "/entrypoint.sh");
--- a/src/itest/java/com/hierynomus/sshj/transport/kex/MLKEMHybridKexIntegrationTest.java
+++ b/src/itest/java/com/hierynomus/sshj/transport/kex/MLKEMHybridKexIntegrationTest.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright (C)2009 - SSHJ Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hierynomus.sshj.transport.kex;
+
+import com.hierynomus.sshj.SshdContainer;
+import com.hierynomus.sshj.SshdContainer.SshdConfigBuilder;
+import net.schmizz.sshj.Config;
+import net.schmizz.sshj.DefaultConfig;
+import net.schmizz.sshj.SSHClient;
+import net.schmizz.sshj.transport.kex.MLKEM768X25519SHA256;
+import net.schmizz.sshj.transport.verification.PromiscuousVerifier;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+import java.util.Collections;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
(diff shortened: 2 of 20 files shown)


exceptionfactory (commented):
Thanks for working on support for this PQC algorithm @alexandrejbr!

As you mentioned in the description, the `MLKEM768` class makes direct use of the Bouncy Castle primitives, as opposed to the JCA.

I have a separate open pull request #1006 that would make Bouncy Castle an optional dependency.

In light of other work to remove explicit use of Bouncy Castle classes, it would be helpful to consider an implementation approach that supports making Bouncy Castle optional.

alexandrejbr:
> Thanks for working on support for this PQC algorithm @alexandrejbr!
> 
> As you mentioned in the description, the `MLKEM768` class makes direct use of the Bouncy Castle primitives, as opposed to the JCA.
> 
> I have a separate open pull request #1006 that would make Bouncy Castle an optional dependency.
> 
> In light of other work to remove explicit use of Bouncy Castle classes, it would be helpful to consider an implementation approach that supports making Bouncy Castle optional.

I can try that route again, but I am quite certain that having a higher java version would be necessary, and from memory it was higher than 21. Do you think that is a reasonable approach @exceptionfactory ?

exceptionfactory:
Thanks for the quick reply @alexandrejbr.

I agree that going with the JCA-based approach could involve issues with Java versions. However, it would be helpful to confirm if it works when Bouncy Castle is registered as a Security Provider.

One possibility is testing for the availability of specific algorithms available to the JCA, which could support an approach based on feature detection.

Along similar lines, another option could be to test for the presence of a specific Bouncy Castle class before registering the new MLKEM implementation factory.

alexandrejbr:
> Thanks for the quick reply @alexandrejbr.
> 
> 
> 
> I agree that going with the JCA-based approach could involve issues with Java versions. However, it would be helpful to confirm if it works when Bouncy Castle is registered as a Security Provider.
> 
> 
> 
> One possibility is testing for the availability of specific algorithms available to the JCA, which could support an approach based on feature detection.
> 
> 
> 
> Along similar lines, another option could be to test for the presence of a specific Bouncy Castle class before registering the new MLKEM implementation factory.

Sounds good @exceptionfactory, I will work on it and report back with an updated implementation.

alexandrejbr:
Hi @exceptionfactory,

The implementation is more complicated than what I would want. I would say there's 2 main problems:

- The JCA keys obtained via SecurityUtils.getKeyPairGenerator("ML-KEM-768").generateKeyPair() include an algorithm identifier while the BouncyCastle MLKEMPublicKeyParameters operates on the raw key directly which is more suitable since that is what is going to be used in the wire format. Managing this difference is a unwanted complexity.

- The other problem is that the decapsulation/encapsulation needed for this hybrid key exchange needs is only part of JDK 21 API (JEP 452) , so in order for this to compile with older version I had to use reflection not only for the usage of BouncyCastle but as well for when JCA is used.

What do you think?

exceptionfactory:
> Hi @exceptionfactory,
> 
> The implementation is more complicated than what I would want. I would say there's 2 main problems:
> 
>     * The JCA keys obtained via SecurityUtils.getKeyPairGenerator("ML-KEM-768").generateKeyPair() include an algorithm identifier while the BouncyCastle MLKEMPublicKeyParameters operates on the raw key directly which is more suitable since that is what is going to be used in the wire format. Managing this difference is a unwanted complexity.
> 
>     * The other problem is that the decapsulation/encapsulation needed for this hybrid key exchange needs is only part of JDK 21 API (JEP 452) , so in order for this to compile with older version I had to use reflection not only for the usage of BouncyCastle but as well for when JCA is used.
> 
> 
> What do you think?

I would defer to @hierynomus as the project maintainer for particular direction, but this does raise some fundamental implementation questions.

Although going in the JCA direction has appeal, the use of reflection adds quite a bit of complexity. Some other JCA-based implementations include similar prefixed algorithm identifiers, so removing those as part of the implementation is not out of the question, as long as it is documented.

The JDK 21 API level seems to be the more problematic concern. Possible solutions include raising the minimum version, although that is more involved. Another possibility is providing this functionality in an independent module, requiring Java 21.

Taking a different path forward, it would also be possible to keep the Bouncy Castle-specific implementation classes, but use Class-availability detection to determine whether to load the corresponding SSHJ classes. This is not trivial, but it is probably cleaner to maintain than reflection for JCA methods.

alexandrejbr:
> > Hi @exceptionfactory,
> 
> > 
> 
> > The implementation is more complicated than what I would want. I would say there's 2 main problems:
> 
> > 
> 
> >     * The JCA keys obtained via SecurityUtils.getKeyPairGenerator("ML-KEM-768").generateKeyPair() include an algorithm identifier while the BouncyCastle MLKEMPublicKeyParameters operates on the raw key directly which is more suitable since that is what is going to be used in the wire format. Managing this difference is a unwanted complexity.
> 
> > 
> 
> >     * The other problem is that the decapsulation/encapsulation needed for this hybrid key exchange needs is only part of JDK 21 API (JEP 452) , so in order for this to compile with older version I had to use reflection not only for the usage of BouncyCastle but as well for when JCA is used.
> 
> > 
> 
> > 
> 
> > What do you think?
> 
> 
> 
> I would defer to @hierynomus as the project maintainer for particular direction, but this does raise some fundamental implementation questions.
> 
> 
> 
> Although going in the JCA direction has appeal, the use of reflection adds quite a bit of complexity. Some other JCA-based implementations include similar prefixed algorithm identifiers, so removing those as part of the implementation is not out of the question, as long as it is documented.
> 
> 
> 
> The JDK 21 API level seems to be the more problematic concern. Possible solutions include raising the minimum version, although that is more involved. Another possibility is providing this functionality in an independent module, requiring Java 21.
> 
> 
> 
> Taking a different path forward, it would also be possible to keep the Bouncy Castle-specific implementation classes, but use Class-availability detection to determine whether to load the corresponding SSHJ classes. This is not trivial, but it is probably cleaner to maintain than reflection for JCA methods.

Thanks for the feedback. I am fine putting the time to get the implementation to where it needs to be. So, some guidance from @hierynomus would be greatly appreciated.

PR #1041, opened 2026-06-03
Add FIDO2/U2F security key (sk-*) support and ssh-agent authentication

### LINKED ISSUE #766: Feature Request: OpenSSH FIDO format support (ed25519-sk and ecdsa-sk key type)
Since v8.2p1 OpenSSH supports FIDO hardware authenticators (e.g. Yubikey) with the key types “ecdsa-sk” and “ed25519-sk", and use of FIDO key is a security requirement in certain situations. 

Is FIDO support on the roadmap for sshj?

Thanks.

squadgazzz:
# Description

A FIDO2 hardware key keeps its private key on the device, so nothing can sign without a physical tap. The usual sshj path of loading a private key and signing in process does not apply. OpenSSH covers this with two key types, `sk-ecdsa-sha2-nistp256@openssh.com` and `sk-ssh-ed25519@openssh.com`, and sshj recognised neither. The connection died before any signing, which blocks every client built on sshj (JetBrains IDEs, DBeaver) the moment a user has a security key.

This adds that support. sshj never touches USB or CTAP, where no mature pure-Java stack exists and Apache MINA sshd stopped short for the same reason. An SSH agent already holds the key and performs the tap, and it returns a finished signature, so sshj speaks the agent protocol and forwards the result. An application that drives the authenticator itself can implement a small signer interface instead.

# Changes

- Add the `sk-ecdsa-sha2-nistp256@openssh.com` and `sk-ssh-ed25519@openssh.com` key types: public-key parsing and serialisation, signature verification, and signature assembly. The verifier rebuilds the authenticator message `SHA256(application) || flags || counter || SHA256(data)` and checks it against the underlying ECDSA or Ed25519 key.
- Add an SSH agent client (`AgentProxy`, `AuthAgent`) over the `SSH_AUTH_SOCK` unix-domain socket. This also gives sshj general ssh-agent authentication for ordinary RSA, ECDSA and Ed25519 keys, which it did not have before. Main sources target Java 8, so the Java 16 socket is reached by reflection, and a caller can supply its own `AgentConnection` on older runtimes or other platforms.
- Add a `SecurityKeySigner` SPI for applications that drive the authenticator directly, and parse `sk-*` private keys from `openssh-key-v1` files.
- Register the new algorithms in `DefaultConfig` and route the security-key signature framing through the publickey auth method.

# How to test

New unit tests. They cover the public-key codec, signature verify and sign round trips, and authentication against an in-process Apache MINA server over both the agent path and the signer path. One test cross-checks sshj's signatures and public keys against MINA's independent `sk-*` implementation, which needs a test-only `net.i2p.crypto:eddsa` dependency.

Verified by hand with a YubiKey: the device produced `sk-ssh-ed25519` signatures with real authenticator flags and counters that the new verifier accepted, and a full login through `AuthAgent` authenticated against a server.

# Related issues

Fixes #766

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -57,6 +57,8 @@ In the `examples` directory, there is a separate Maven project that shows how th
 
 * reading known_hosts files for host key verification
 * publickey, password and keyboard-interactive authentication
+* FIDO/U2F security keys (`sk-ecdsa-sha2-nistp256@openssh.com`, `sk-ssh-ed25519@openssh.com`), via an SSH agent or a pluggable `SecurityKeySigner`
+* SSH agent authentication over the `SSH_AUTH_SOCK` unix-domain socket
 * command, subsystem and shell channels
 * local and remote port forwarding
 * scp + complete sftp version 0-3 implementation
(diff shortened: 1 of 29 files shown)


squadgazzz:
Split into two focused PRs for easier review: #1042 (ssh-agent authentication) and #1043 (FIDO2 `sk-*` key support). Together they deliver YubiKey auth over the agent.

PR #1043, opened 2026-06-03
Add FIDO2/U2F security key (sk-*) support

### LINKED ISSUE #766: Feature Request: OpenSSH FIDO format support (ed25519-sk and ecdsa-sk key type)

squadgazzz:
# Description

A FIDO2 hardware key keeps its private key on the device, so nothing can sign without a physical tap. The usual sshj path of loading a private key and signing in process does not apply. OpenSSH covers this with two key types, `sk-ecdsa-sha2-nistp256@openssh.com` and `sk-ssh-ed25519@openssh.com`, and sshj recognised neither, so a connection died before any signing. That blocks every client built on sshj (JetBrains IDEs, DBeaver) the moment a user has a security key.

This adds the key types, their signature verification, and a way to sign. sshj never touches USB or CTAP, where no mature pure-Java stack exists and Apache MINA sshd stopped short for the same reason. An application implements a small `SecurityKeySigner` to reach the authenticator, or it lets ssh-agent do the work (agent support is in #1042).

# Changes

- Add the `sk-ecdsa-sha2-nistp256@openssh.com` and `sk-ssh-ed25519@openssh.com` key types: public-key parsing and serialisation, and signature verification. The verifier rebuilds the authenticator message `SHA256(application) || flags || counter || SHA256(data)` and checks it against the underlying ECDSA or Ed25519 key.
- Add a `SecurityKeySigner` SPI and assemble the security-key signature, which carries the raw signature plus the authenticator flags and counter, through the `publickey` method.
- Parse `sk-*` private keys from `openssh-key-v1` files.
- Register the new algorithms in `DefaultConfig`.

# How to test

New unit tests. They cover the public-key codec, signature verify and sign round trips, and authentication against an in-process Apache MINA server through the `publickey` method. One test cross-checks sshj's signatures and public keys against MINA's independent `sk-*` implementation, which needs a test-only `net.i2p.crypto:eddsa` dependency.

Verified by hand with a YubiKey: the device produced `sk-ssh-ed25519` signatures with real authenticator flags and counters that the new verifier accepted.

# Related issues

Fixes #766. Pairs with the ssh-agent support in #1042, the usual way a security key authenticates.

This PR and #1042 both edit `README` and `KeyedAuthMethod`, so whichever merges second needs a trivial rebase.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -57,6 +57,7 @@ In the `examples` directory, there is a separate Maven project that shows how th
 
 * reading known_hosts files for host key verification
 * publickey, password and keyboard-interactive authentication
+* FIDO/U2F security keys (`sk-ecdsa-sha2-nistp256@openssh.com`, `sk-ssh-ed25519@openssh.com`) via a pluggable `SecurityKeySigner`
 * command, subsystem and shell channels
 * local and remote port forwarding
 * scp + complete sftp version 0-3 implementation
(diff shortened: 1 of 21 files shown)


hierynomus:
The only "potential" problem with the PR is that net.crypto.eddsa introduces a vulnerability. Is there a way around that library?

squadgazzz:
Done in 6529c7a. It was test-scope only, never a runtime dependency. I had pulled it in so Apache MINA's `sk-ssh-ed25519` verifier could independently cross-check our signatures, but that verifier delegates to `net.i2p.crypto.eddsa`. Since sshj dropped eddsa in 0.39.0, I took it back out along with the two checks that relied on MINA's ed25519 path (the interop test and one `AuthSecurityKeyTest` case).

`sk-ssh-ed25519` stays covered without it: `SecurityKeySignatureTest` builds the authenticator blob per PROTOCOL.u2f and round-trips sign and verify, and the MINA cross-check still runs for `sk-ecdsa`. The auth-layer signature framing is identical for both key types, so the ecdsa end-to-end case exercises it. I also confirmed `sk-ssh-ed25519` against a real YubiKey out of band. No eddsa anywhere in the build now.

squadgazzz:
@hierynomus, please let me know if there is anything else I should change.

hierynomus (approved):

Issue #1053, opened 2026-09-07
PKCS8KeyFile cannot load PBES1-encrypted PKCS#8 keys since 0.40.0 — NoSuchAlgorithmException from AlgorithmParameters.toString()


ansokpl:
### Summary

Since 0.40.0, `loadKeys(path, passphrase)` fails for a passphrase-protected PKCS#8 key that
uses **PBES1** encryption (e.g. `PBEWithSHA1AndDESede`, `PBEWithMD5AndDES`). The same key loads
correctly with 0.39.0, so this is a regression introduced by the `PKCS8KeyFile` rewrite.

PBES2 keys (the OpenSSL default, e.g. `-aes-256-cbc`) still work — but, as shown below, only by
accident.

### Version

* sshj **0.40.0** (works on 0.39.0)
* JDK 25 (Temurin), Linux
* Reproduced both with and without BouncyCastle (`bcprov`/`bcpkix` 1.85) on the classpath

### Steps to reproduce

```bash
openssl genpkey -algorithm RSA -pkeyopt rsa_keygen_bits:2048 -out key.pem
openssl pkcs8 -topk8 -v1 PBE-SHA1-3DES -in key.pem -passout pass:test-passphrase -out key_pbes1.pem
```

```java
KeyProvider keyProvider = new SSHClient().loadKeys("key_pbes1.pem", "test-passphrase".toCharArray());
System.out.println(keyProvider.getPublic());   // throws
```

### Actual result

```
java.io.IOException: PEM Key [PKCS8] processing failed
    at net.schmizz.sshj.userauth.keyprovider.PKCS8KeyFile.readEncryptedPkcs8KeyPair(PKCS8KeyFile.java:248)
    at net.schmizz.sshj.userauth.keyprovider.PKCS8KeyFile.readKeyPair(PKCS8KeyFile.java:123)
    at net.schmizz.sshj.userauth.keyprovider.PKCS8KeyFile.readKeyPair(PKCS8KeyFile.java:106)
    at net.schmizz.sshj.userauth.keyprovider.BaseFileKeyProvider.getPublic(BaseFileKeyProvider.java:90)
Caused by: java.security.NoSuchAlgorithmException:
    salt:
[0000: 2E 28 18 CB 47 51 15 FD                            .(..GQ..
]
    iterationCount:
    0800
 SecretKeyFactory not available
    at java.base/javax.crypto.SecretKeyFactory.getInstance(SecretKeyFactory.java:164)
[...]

PR #1055, opened 2026-09-09
Fix loading PBES1 encrypted PKCS8 private keys

exceptionfactory:
This pull request corrects handling of PKCS8 private keys encrypted using the historical [PBES1](https://www.rfc-editor.org/info/rfc8018/#section-6.1) strategy described in issue #1053.

Refactoring to reduce the use of Bouncy Castle for reading encrypted PKCS8 private keys in pull request #989 introduced an approach that relied on [AlgorithmParameters.toString()](https://docs.oracle.com/javase/8/docs/api/java/security/AlgorithmParameters.html) to return the required algorithm name for the `SecretKeyFactory`. This approach works for the newer [PBES2](https://www.rfc-editor.org/info/rfc8018/#section-6.2) strategy, but returns a serialized representation of the binary structure when the strategy is a variant of `PBES1`. This implementation worked with available test files.

The revised approach uses [AlgorithmParameters.getAlgorithm()](https://docs.oracle.com/javase/8/docs/api/java/security/AlgorithmParameters.html#getAlgorithm--) to determine whether the strategy matches the general `PBES2` algorithm, requiring the `toString()` method, and otherwise falls back to the value returned from `getAlgorithm()`.

The fallback strategy avoids reverting to Bouncy Castle classes and includes a new test method with a test file generated using `openssl` as described in issue #1053.


hierynomus (approved):

hierynomus:
Thanks for the quick PR @exceptionfactory!

